PR #42, opened 2015-02-27
add methods to get Earth latlon for Apparents and EarthSatellites

trehn:
This is the solution to [my question at StackOverflow](http://stackoverflow.com/questions/28707632/getting-latitude-and-longitude-of-the-sun-on-a-world-map-with-pyephem) on how to get the Earth lat/lon position of the Sun (or other bodies) so they can be shown on a world map. It's also meant to fix [pyephem/#16](https://github.com/brandon-rhodes/pyephem/issues/16).

The actual math is straight out of the version of libastro that's bundled in PyEphem.

I did my best to integrate this with Skyfield's classes so I hope I put things in the right place.

Unfortunately, this will only work for Earth. I have no idea how to tweak the magic numbers in the code to work for other bodies.

### DIFF
--- a/skyfield/documentation/earth-satellites.rst
+++ b/skyfield/documentation/earth-satellites.rst
@@ -98,7 +98,7 @@ satellite is flying:
 
 .. testcode::
 
-    topos = sat.over_topos(JulianDate(utc=tup))
+    topos = sat.over_location(JulianDate(utc=tup))
 
     print(topos.latitude)
     print(topos.longitude)
--- a/skyfield/documentation/earth-satellites.rst
+++ b/skyfield/documentation/earth-satellites.rst
@@ -93,6 +93,21 @@ See :doc:`positions` to learn more about these possibilities:
     01h 55m 39.18s
     +62deg 55' 57.4"
 
+Finally, you can find the point on the Earth's surface over which the
+satellite is flying:
+
+.. testcode::
+
+    topos = sat.over_topos(JulianDate(utc=tup))
+
+    print(topos.latitude)
+    print(topos.longitude)
+
+.. testoutput::
+
+    51deg 20' 12.0"
+    -84deg 09' 38.5"
+
 The standard SGP4 theory of satellite motion that Skyfield uses
 is a rough enough model of the near-Earth environment
 that it can only predict a satellite position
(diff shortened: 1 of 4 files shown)


brandon-rhodes:
(Returning to the project after a few years of other obligations —) This pull request is so well assembled, with documentation and everything, that I'm going to go ahead and land it even though the math is a bit of a problem: that old technique for computing lat and lon isn't going to match the higher accuracy WGS84 coordinates that Skyfield returns. But after landing I'll test it, by choosing a location above a given Topos and seeing if this method gives me the Topos's own lat and lon back.

Thanks for this very nice code, though, and my going ahead and merging will force me to hunt down the correct math and get it in the next version of Skyfield. Thanks, and I'll try to remember to comment back here about what I find!

brandon-rhodes:
I've merged the main commit with the logic, and so I'm gratefully closing this pull request — thanks for the impetus to get this in master!

I am likely to make one change to the API: I am wary about the fact that this makes it necessary for `positionlib.py` to import `toposlib.py`, because that puts Skyfield halfway to having an import loop; all that would need to happen would be a future need for `toposlib.py` to need to create a position, and we'd be stuck. So I've asked myself the question, Q: why would the position library need to import `Topos`? A: so that it can create one — because, I suddenly realized, the idea of putting `over_topos()` methods everywhere is not only repetitive (there might be several kinds of objects that need it), but it essentially strands what's really a Topos constructor over in another class.

So I think I'm going to pivot `over_position()` to be something that a Topos should know how to do, not that a random position should have the expertise to do. I'll make it a class method of Topos and maybe name it `Topos.beneath(...)` that takes another position as its argument. The code can then live in `toposlib.py`, which feels like where it belongs, since the numeric routine is specific to locations on the Earth's surface.

Thanks for getting this feature started, I look forward to it being in the next version of Skyfield!

brandon-rhodes:
(Oh, and, another reason I'll be deleting and replacing the actual computational code, besides the fact it doesn't give the same results as the forward code: PyEphem is released under a more restrictive license than Skyfield, so technically I can't now continue to release Skyfield under its traditional MIT license until the code is removed.)

brandon-rhodes:
Oh, and, for clarity I should illustrate why we should turn xyz into latlon using WSG84. Otherwise, going from latlon → xyz → latlon will return a different coordinate than the user started with. Here's an example:

```
>>> print(Topos.beneath(Topos(latitude_degrees=45, longitude_degrees=0).at(t)))
Topos 44deg 42' 52.6" N 00deg 00' 29.6" E
```

PR #43, opened 2015-02-28
[WIP] more attributes for EarthSatellites

trehn:
Note that this depends on my other PR, #42.

I have [a whole lot of other stuff](https://github.com/trehn/termtrack/blob/master/termtrack/satellite.py#L70) that I would love to move into Skyfield. Before I do that, however, I'd like some feedback on what I've done so far and if you're at all interested in having this here or if it's out of scope.

Also, as I add more methods and attributes it will become very inefficient to pass a jd to each of them separately. IMO `EarthSatellite.__init__()` should also take a jd and then all methods can just use that. Or we could go with a `compute()` method, PyEphem-style.

Please let me know what you think. And thanks for all the work you're doing in this area!


brandon-rhodes:
Thank you for both of these pull requests — I have been so busy this last week trying to get Earth locations working with the new code for using SPICE ephemeris SPK files, that I was not visiting GitHub often enough to see your contributions!

Later this weekend, I hope to get the chance to merge your contributions, go through them making tweaks to match the conventions I want in Skyfield code, then I'll come back and comment to let you know how to tweak any future contributions, so that we can land the rest of your ideas.

Thanks so much for making the effort to contribute back, instead of just keeping your changes private and never getting around to letting the rest of us enjoy them! I'll try to get to this soon.

trehn:
Great, thank you.

In retrospect I think it would be better to rename ``over_topos()`` to something more intuitive like ``over_location()``. I'm a friend of verbose and clear names for variables and methods.

Please try and look into the changes I mentioned in the second half of the original comment about ``compute()`` (or something similar). So many of these values depend on each other that I really believe there is no way around a change like that if I am to add more attributes.

trehn:
(Travis just failed on 2.6 because of the set comprehensions you removed after I created the branch for this PR, so everything is fine)

brandon-rhodes:
Having spent all of my spare time this week getting the new `Topos` object's first tests passing — I finally figured out that what HORIZONS calls `Delta-t` is not what astronomers traditionally call ΔT — the days are now counting down towards PyCon, and my attention needs to turn to finishing my talk and tutorial.

While that will take me away from Skyfield for three weeks, I have decided to devote my entire time at the PyCon sprints, the week following the conference, to giving you feedback, working to make your additions to the code match what's already there, and getting Skyfield ready for a new release. Much better to wait until the sprints, I think, than for me to try to rush something out the door right now and then not be happy with it.

(Any chance you are attending PyCon, or will be at the sprints? I guess you probably find conferences in Europe far more convenient and reasonable!)

But so that you are not left waiting for three weeks with nothing to do, I will try to comment on your pull requests this weekend so you have my first thoughts on your work. Thanks again!

skyfield/constants.py
```diff
@@ -22,12 +22,15 @@
 ANGVEL = 7.2921150e-5
 ERAD = 6378136.6
 IERS_2010_INVERSE_EARTH_FLATTENING = 298.25642
+IERS_2010_EARTH_FLATTENING = 1 / IERS_2010_INVERSE_EARTH_FLATTENING
```
brandon-rhodes: What was the motivation for moving this out of `earthlib.py` — do you have plans to use it somewhere else as well, maybe in the SGP4 code?

skyfield/documentation/positions.rst
```diff
@@ -39,7 +39,8 @@ together with all of the attributes and methods that they support:
 
     Apparent position only
      │
-     └── `altaz(…) <api.html#Position.altaz>`_            →   alt, az, distance
+     └── `altaz(…) <api.html#Position.altaz>`_                 →   alt, az, distance
+     └── `over_location(…) <api.html#Apparent.over_location>`_ →   topos
```
brandon-rhodes: I like how you added this in a way that is so symmetric with the rest of the docs! I will have to think about the name a bit, though; I have normally heard this referred to using the “sub” prefix, like the “sublatitude” and “sublongitude” of the satellite, and HORIZONS uses standard phrases like “Sun sub-longitude & sub-latitude” and “Sub-Sun position angle & distance” and “Observer sub-lon & sub-lat” — so I will probably be tempted to switch the language around to match those existing conventions that astronomers are familiar with, but I'll think about it for a few weeks and decide at PyCon. :)

skyfield/documentation/positions.rst
```diff
@@ -98,6 +99,10 @@ as its argument and return a corresponding number of positions.
     astrometric = boston(jd).observe(mars)
     apparent = boston(jd).observe(mars).apparent()
 
+    # Earth location where in zenith at date
+
+    topos = apparent.over_location()
```
brandon-rhodes: Wow! You return a new `topos` object! That would never had occurred to me — I just would have returned a `lat, lon` pair from a `sublatlon` function or something, which would have been a dead end because you can't ask a bare `lat` and `lon` further questions like you can use a `Topos` to follow up. I think I like the idea, and I think we will definitely keep this approach!

skyfield/positionlib.py
```diff
@@ -329,3 +344,27 @@ def ITRF_to_GCRS(jd, rITRF):  # todo: velocity
...
+    x, y, z = xyz
+
+    # the following block is ported from GetSubSatPoint() in
+    # earthsat.c of libastro 3.7.6
+    sidereal_solar = 1.0027379093
+    sid_day = floor(dublin_julian_date)
+    t = (sid_day - 0.5) / 36525
+    sid_reference = (6.6460656 + (2400.051262 * t) +
+                     (0.00002581 * (t**2))) / 24
+    sid_reference -= floor(sid_reference)
```
brandon-rhodes: I am not sure — I may not have ever read this PyEphem code before — but this might be trying to compute, using an old and inexact formula, the sidereal time for the Julian Date in question? If so, you might want to take a look at the attribute `jd.gast` on the Julian Date object, which uses a modern IAU formula for the same thing. Try adding some `print` statements here that compare the numbers this old PyEphem formula is generating with the number that `.gast` produces and let me know if you see any resemblance. Thanks!

skyfield/sgp4lib.py
```diff
@@ -87,6 +103,41 @@ def gcrs(self, jd):
...
+        >>> from skyfield.api import earth, JulianDate
+        >>> tle = ("ISS (ZARYA)\n"
+        ... "1 25544U 98067A   15058.48161588  .00023857  00000-0  35618-3 0  9991\n"
+        ... "2 25544  51.6478 271.9610 0008043  54.4849  18.6646 15.54887163930964")
+        >>> iss = earth.satellite(tle)
+        >>> topos = iss.over_location(JulianDate(utc=(2015, 2, 27, 22, 22, 0)))
+        >>> topos.latitude
+        <Angle 49deg 19' 48.3">
+        >>> topos.longitude
+        <Angle -155deg 37' 41.8">
```
brandon-rhodes: I would like to see this test go one step farther: now that you have a `Topos`, ask it to observe the satellite, and demonstrate that the `Topos` thinks that (a) the satellite has an altitude of exactly (or very nearly) 90° above the horizon, and (b) states the range to the satellite is the same as the elevation that is returned by your elevation function. You can get the altitude angle and range by running `.altaz()`, I think. By doing a full “round trip” from the satellite to the `Topos` and then back the other way, we will learn whether all the math is consistent. Thanks!

skyfield/sgp4lib.py
```diff
@@ -100,6 +151,10 @@ def _observe_from_bcrs(self, observer):
         # g.distance = euclidian_distance
         return g
 
+    @property
+    def semi_major_axis(self):
+        return Distance(m=(GE / (self.mean_motion ** 2)) ** (1/3))
```
brandon-rhodes: This formula brings up a big issue that at some point we should tackle. (But maybe it's too big an issue to look at right now, and we could circle back once these first pull requests are merged?) The issue is: that the semi-major axis is already computed in the SGP4 library, but it might not get saved to the `satrec` object in a way that we can use.

The code is a mess because I just barely changed the C into Python, but here is the computation of the semi-major axis that this satellite object will already have done over in the `sgp4` library:

https://github.com/brandon-rhodes/python-sgp4/blob/master/sgp4/propagation.py#L1088

https://github.com/brandon-rhodes/python-sgp4/blob/master/sgp4/propagation.py#L1152

On the one hand, it looks like several of the features and methods and attributes you want to add could be added all the way down at the `sgp4` library level, and make it a much nicer library even for people who want to use it alone without Skyfield.

On the other hand, it is such terrible code that I maybe cannot ask you to read it and make a pull request against it for things like the semi-major axis. :)

But it would be very satisfying to be able to add as many features to `sgp4` itself, and then simply expose them through Skyfield, instead of doing satellite math up in the Skyfield code instead of letting the satellite code take care of it.

Take a look, and let me know what you think — thanks! :)

skyfield/constants.py
trehn: I don't think I need it anywhere else right now, just wanted to keep the constants in one place.

skyfield/documentation/positions.rst
trehn: Having no background in astronomy whatsoever, I chose a name that is more intuitive for people like me. While I like to keep things as dumb as possible, the primary audience of Skyfield probably knows about things like this and we should cater to them.

trehn:
Thank you for taking the time to review this! I'd love to be at PyCon, but I can't make it. I'm starting at a new job on Monday, so my time will be limited during the next few weeks as well while I settle in.

I'll definitely look into the things I didn't reply to as soon as I can.

skyfield/constants.py
brandon-rhodes: Okay! We will leave it there for now.

skyfield/documentation/positions.rst
brandon-rhodes: You are entirely correct that names are important, and I am indeed wanting people new to astronomy to be a primary audience for Skyfield. But, sometimes, if we can find a more “official” name for a concept, then the Skyfield user — while thinking they are just writing a quick program — is also learning official terminology that will help them out when the look at other programs and web sites. The name does not stand in the way of the pull request being merged, so we can keep it this way until I have time at PyCon to look into what other libraries call this idea. Thanks!

brandon-rhodes:
No rush — if you can try out the `gast` idea and the round-trip within the next 4 weeks, then you will have the patch ready for when the sprints start at PyCon. Thanks!

skyfield/documentation/positions.rst
leblancfg: Would like to cast +1 for `sub-point` instead of `over_location`. See https://celestrak.com/columns/v02n03/.

leblancfg:
Is this PR still being considered? Am willing to put in some hours towards this so that it gets pulled in.

brandon-rhodes:
Yes, I am still open to a richer attribute experience for Skyfield Earth satellites! Take a look over the comments in the above conversation, and see if you can incorporate them into your diff — and I'll watch for the update.

dmopalmer:
With a two-year bump:

For Earth satellites, I want to implement:
1. next_pass() type information, when and where it will rise,culminate,set
2. TLE output

Will this pull request be pulled, and therefore I should build on it, or should I fork from the master?

brandon-rhodes:
@dmopalmer I'd recommend working atop `master`. See the `https://rhodesmill.org/skyfield/api-position.html#skyfield.positionlib.Geocentric.subpoint` routine for where a similar idea to this PR landed in the code in the meantime. Also, see the “almanac” routines for rise / set functions that maybe could be called for use with satellites.

brandon-rhodes:
I'm going to temporarily close this to keep my dashboard clear, but I'm open to further rounds of discussion! In the meantime, `.subpoint()` is now fully part of Skyfield and works for all kinds of objects not just Earth satellites (in case anyone has been following this wanting that specific feature). And it returns a Topos just like this PR suggested.

Feel free to ask about other aspects of this PR that you might want to work further on, and I can comment on the form in which those improvements would be most likely to be merged. Thanks!

PR #44, opened 2015-03-07
cache EarthSatellite methods by JulianDate

trehn:
Sorry for the spam. This is a possible solution for the problem I mentioned in #43: Adding more attributes to EarthSatellites will result in lots of repeated calls to methods like `gcrs()` since the formula behind a particular attribute may depend on other attributes.

For example:

Mean anomaly can be computed from just the jd and "constants" provided by the TLE data. Thus, there is a method called `mean_anomaly(self, jd)`. To compute the eccentric anomaly, I need mean anomaly and eccentricity, resulting in a method like this:

```
def eccentric_anomaly(jd):
    return compute_eccentric_anomaly(self.mean_anomaly(jd), self.eccentricity)
```

Now to get both eccentric anomaly and mean anomaly, a user would need to call these methods like this:

```
m = sat.mean_anomaly(jd)
e = sat.eccentric_anomaly(jd)
```

This is inefficient because the second call will compute mean anomaly again. The problem is compounded because of dependencies like this:
- `time_since_periapsis` depends on `mean_anomaly`
- `time_to_periapsis` depends on `time_since_periapsis`
- `true_anomaly` depends on `eccentric_anomaly`

The decorator implemented here will cache computed results until a different JulianDate is given. This would solve the problem without changing the API (like adding a `compute()` method as I suggested previously, which might also compute more values than the user wanted).

If you agree with this, you should probably merge it before #42 and #43. I can then rebase those so they'll apply cleanly.

### DIFF
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -8,7 +8,7 @@ from sgp4.propagation import sgp4
 from .constants import AU_KM, DAY_S, T0, tau
 from .functions import rot_x, rot_y, rot_z
 from .positionlib import Apparent, Geocentric, ITRF_to_GCRS
-from .timelib import JulianDate, takes_julian_date
+from .timelib import cache_for_julian_date, JulianDate, takes_julian_date
 
 # important ones:
 # jdsatepoch
@@ -27,6 +27,8 @@ class EarthSatellite(object):
 
     def __init__(self, lines, earth):
         sat = twoline2rv(*lines[-2:], whichconst=wgs72)
+        self._cached_jd = None
+        self._cached_jd_values = {}
         self._sgp4_satellite = sat
         self._earth = earth
         self.epoch = JulianDate(utc=(sat.epochyr, 1, sat.epochdays - 1.0))
@@ -76,6 +78,7 @@ class EarthSatellite(object):
         return rGCRS, vGCRS, error
 
     @takes_julian_date
+    @cache_for_julian_date
     def gcrs(self, jd):
         """Return a GCRS position for this Earth satellite.
 
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -49,6 +49,24 @@ extra_documentation = """
 
 """
 
+def cache_for_julian_date(method):
+    """Decorator that can be applied to methods that take a JulianDate
+    as first argument. Will cache method return value for any subsequent
+    calls with the same JulianDate.
+    """
+    def cached_method(self, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


brandon-rhodes:
I will look at this again later this weekend, but I suspect that the problem of caching might go away if we move these attributes on to the Apparent class instead, where the instances already holds a specific date and precisely an object's relative position to the observe on that date. Then these methods could just look at `self.jd` and the `self.position` vector, along with the observer's orientation that is also available, and compute these attributes.

If we thought the user might look at them many times, we could add them to `__getitem__()` like I have done with expensive items on the JulianDate, but only if we can't otherwise figure out how to make them available simply.

I will comment again later in the weekend after thinking about it some more, though!

trehn:
Mh, I don't believe Apparent would be a good place. A position in the sky ist just that, attaching stuff like like orbital velocity to it seems wrong to me. Those are properties of a satellite. Besides, most of this stuff is available only for EarthSatellites, not the other bodies for which Apparents can be computed.

brandon-rhodes:
Position objects all have a slot for both the x-y-z vector and the dx-dy-dz velocity vector:

    https://github.com/brandon-rhodes/python-skyfield/blob/master/skyfield/positionlib.py#L30

So I think that velocity is indeed part of a position — all of the NASA and USNO ephemeris tools, I have found, consider position and velocity to together both be part of where something is. The `sky_pos` structure returned by the C-language NOVAS library, to take one quick example from a non-Python library, includes a velocity in the `sky_pos` struct that its central `place()` function returns:

https://github.com/brandon-rhodes/python-novas/blob/master/Cdist/novas.h#L184

But you do have an excellent point when you note that “most of this stuff is available only for EarthSatellites” — if we stuck these attributes directly on Astrometric or Apparent, they would be useless most of the time. It is likely that we need a position subclass specifically for Earth satellites that bundles together all the information about a satellite at a given `jd`.

brandon-rhodes:
Apparently the 2015 version of myself had a very hard time saying “no” to making code more complicated to handle special cases, and left this pull request dangling, which in retrospect I don’t think was quite fair. To express clearly what I probably only vaguely felt at the time: I prefer to maintain software with as little hidden behavior as possible, and this particular optimization handles a narrow enough case that I am not inclined to bring it into the main codebase.

Another possible way to solve this issue, if interest again arose, would be to write a single routine "get satellite orbital data" that returns a single struct listing all the parameters, each of which would be computed once. That would make clear that "all this stuff should be computed in a single step", without any magic. If only all Pythons back to 2.6 had data classes, we'd even have a pretty way to wrap up the set of returned values!

PR #45, opened 2015-03-07
remove dict comprehensions for Python 2.6

trehn:
to get rid of the Travis failures on the other PRs

### DIFF
--- a/skyfield/chaining.py
+++ b/skyfield/chaining.py
@@ -22,7 +22,7 @@ class Body(object):
 
     def observe(self, body):
         every = self.segments + body.segments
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Great — thanks for getting the build fixed on this PR!

Once your code is merged I will want some time to play with the new API calls and give you feedback, which means that we want a long runway until the next release. So my plan is to document and get a new Skyfield out the door this weekend or early next week. That will then let us land this PR with plenty of time before the next release during which I can try out the new API and give you feedback — and so that you can get more experience with it as well and decide what you think the best method names are and so forth.

PR #47, opened 2015-04-13
Updating requirements file with missing imports: jplephem and de421

aarose:


tritium21:
(I am not at PyCon, so if this was discussed in meat space, I don't know)

Are those actually required to hack on skyfield and/or run the tests?  Requirements.txt should be for the minimum required to effectively hack on the code.  since all the ephemerides are the same for that purpose, do you need more than one in the requirements?

brandon-rhodes:
These requirements were already listed in `setup.py` because they are hard requirements of the project itself — whereas `requirements.txt`, before this change, had only the *additional* requirements that are needed for development and testing from inside the repo.

But some of our sprinters are not in the habit of doing a `pip install -e` of a project, and instead expect to have `requirements.txt`list *everything* needed for development, including the project's hard dependencies themselves. To get these developers going quickly, we added them.

Why do we need several ephemerides to develop? Good question! We need several ephemerides for testing because different sources of test numbers (the USNO, the JPL) use different ephemerides to produce their answers that we want to check against.

PR #69, opened 2015-11-09
added iosurvey module to compare file headers from urls

aerler:
This is Andre from the PyCon sprint with the little header comparison tool.

### DIFF
--- a/skyfield/iosurvey.py
+++ b/skyfield/iosurvey.py
@@ -1,0 +1,29 @@
+"""Routines to explore available datasets
+"""
+
+# load low-level file tools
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thank you! I will take a look at this and see about getting it merged.

brandon-rhodes:
Thank you! I was happy to have you at the sprint and will be looking at this code soon to incorporate it better into the Skyfield toolset. I have just committed it to the master branch, so it is now a permanent part of the project history!

aerler:
OK, cool! It was a pleasure to work with you!

PR #75, opened 2015-12-08
Use `is` operator for comparing with `None` (Pep8)

proinsias:
I'm trying out the new [Quantified Code](https://www.quantifiedcode.com/) autofixes, and it suggested this change.


brandon-rhodes:
I have just committed this to master. Thanks! :)

PR #77, opened 2016-01-09
Add conda build config.

jochym:
I have created packaging setup for anaconda/conda system based on my conda packages of skyfield (https://anaconda.org/jochym/skyfield). I think it will be beneficial to have this in the upstream package. I can keep building the packages from my tree for now but eventually it would be better to have it under the upstream command.


brandon-rhodes:
I was recently able to delete the conda rules from my `jplephem` project because, if I understand correctly, there now exists a “feedstock” project that provides a home for conda rules. Would it be possible for the rules to live there, where they could be kept up to date and maintained as needed by the conda folks, instead of having to live here in a repository that is not conda-specific? Here is a link to the rules they created for that other project:

https://github.com/conda-forge/jplephem-feedstock

brandon-rhodes:
The `conda-forge` project seems to be continuing to gain momentum, so I am going to encourage tool-specific files like this to go there, instead of having the main repository ballooned by install formulae from third-party tools. If I reconsider later than I'll re-open this pull request, but for the moment I think that these files will find a happier home over in the conda ecosystem!

jochym:
@brandon-rhodes : May I add you to the maintainer list for your packages?

brandon-rhodes:
Yes, though I've not learned the details of conda packaging and might not be much help if there were an actual bug!

PR #92, opened 2016-05-13
fix print spelling

SeanBE:


brandon-rhodes:
Using `print` there would put all of the elements on the same line, making a very long line in the documentation. I chose pprint https://docs.python.org/2/library/pprint.html because it splits the long lines into several lines by putting each element on its own line.

SeanBE:
Oh. My bad! Thanks for the library.

brandon-rhodes:
No problem! Thanks for the attempt, and I hope you find Skyfield useful!

Issue #97, opened 2016-07-27
skyfield.api circular import crash


jrs65:
In commit d8d41345 it seems that a circular import was introduced that causes a crash when`skyfield.api` is imported.

```python
>>> from skyfield import api
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "skyfield/api.py", line 18, in <module>
    from .named_stars import NamedStar
  File "skyfield/named_stars.py", line 5, in <module>
    from .data import hipparcos
  File "skyfield/data/hipparcos.py", line 2, in <module>
    from skyfield import api
ImportError: cannot import name api
```

A simple fix is to move the import into the `skyfield.data.hipparcos.load` routine that needs it.

I'll try and submit a PR with this later today.

brandon-rhodes:
Oddly enough, I can't reproduce this — what Python version are you using? Could that be the difference between my setup and yours?

jrs65:
Oh, sorry, forgot to say. This is in Python 2.7.10, on OS X 10.11.6. I can imagine things have changed significantly in Python 3, if that's what you are mostly using.

jrs65:
Just to confirm, it seems to work fine on Python 3.5. I only have the issue on 2.7.

brandon-rhodes:
Thanks for contributing 176fdef7df325defc6cfa789e65261b71b84eafb which fixed this!

PR #100, opened 2016-07-28
Quick fix for the circular import issue

jrs65:
This is a quick fix for #97 which is currently preventing skyfield being used.


brandon-rhodes:
Thank you! I merged this in 176fdef7df325defc6cfa789e65261b71b84eafb.

Issue #116, opened 2016-10-21
Satellite reading extremely slow due to loading timescale


dmopalmer:
I load in a file of TLEs and convert to Earth satellites, and it takes 5 minutes (vs 0.2 s in pyephem).

Profiling, I find that you are reloading the timescale for every single satellite.

sgp4lib.py:
```
class EarthSatellite(object):
    def __init__(self, lines, earth):
        ....
        self.epoch = api.load.timescale().utc(sat.epochyr, 1, sat.epochdays)
```

I suggest that the timescale should be cached as a class variable.  (By lazy evaluation so that skyfield.api.load can be repointed to the relevant directory before the first time it is used and so that if you never use EarthSatellite it doesn't download and drop the loaded leapseconds file in the current directory.)

brandon-rhodes:
Thanks for the pull request!

PR #117, opened 2016-10-21
Issue #116: EarthSatellite now caches the timescale 

dmopalmer:
`EarthSatellite.__init__()` now caches the timescale as a class variable Instead of loading it each time.

This increases the speed of the initialization by ~100x (from ~20 ms to ~0.23 ms), which is relevant when loading complete satellite catalogs (3 minutes to 2 seconds).

This is 20x slower than pyephem, but fast enough that the speed difference is usually not relevant.

This is the fix to issue #116 .  If the same code is in the stand-alone sgp4 library it should be propagated there as well.

### DIFF
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -24,15 +24,18 @@ _minutes_per_day = 1440.
 
 class EarthSatellite(object):
     """An Earth satellite loaded from a TLE file and propagated with SGP4."""
+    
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thank you! I am likely to solve the problem slightly differently for the next full release, but for now I'll merge this so that people working on `master` get better performance — they will be very happy with this change!

Issue #95, opened 2016-06-06
Skyfield should offer a way to discover DE431’s filename


davidmikolas:
I wanted to resolve a sub-kilometer difference between Skyfield and JPL Horizons, so I thought I would load DE431, since my Horizons output lists this as the ephemeris used.

Here's what I got on the terminal.  DE421 loaded in a few seconds, DE431 gives a url open error. 

```
>>> data = load('de421.bsp')
[#################################] 100% de421.bsp
>>> data = load('de431.bsp')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/uhoh/anaconda/envs/npy/lib/python2.7/site-packages/skyfield/iokit.py", line 71, in __call__
    return load(filename, self.directory, verbose=self.verbose)
  File "/Users/uhoh/anaconda/envs/npy/lib/python2.7/site-packages/skyfield/iokit.py", line 176, in load
    download(url, path, verbose=verbose)
  File "/Users/uhoh/anaconda/envs/npy/lib/python2.7/site-packages/skyfield/iokit.py", line 207, in download
    raise IOError('cannot get {0} because {1}'.format(url, e))
IOError: cannot get ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de431.bsp because <urlopen error ftp error: [Errno ftp error] 550 Failed to change directory.>
```

brandon-rhodes:
NASA has chosen the filename `de431t.bsp` for that ephemeris:

ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/

I will edit this issue to add the idea that Skyfield should maybe offer a way to list ephemerides, or even to let the user search them given a substring to search for.

uhoh-SE:
Great! A simple method that returned a list or dict would be excellent (and pythonic), with size, begin and end dates available. Maybe a text/comment field with the gory details, but that's icing.

Something along the lines of:

    infodict = load('info')

brandon-rhodes:
Running a simple list-directory on each external source of ephemerides would be enough to collect the filenames. For the beginning and ending dates, would we pull the first few blocks of the `.bsp` file to read its header information?

brandon-rhodes:
[...]

PR #134, opened 2017-05-09
Added list_bsp method to load (issue #95)

bwgref:
This addresses issue #95. It adds the list_bsp method to load, which queries the remote URLs to figure out what bsp files are available. Just does a simple thing and assumes the touch date is the when the thing was generated and leaves it up to the user to figure out what they should actually use.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -236,9 +236,66 @@ class Loader(object):
         leap_dates, leap_offsets = self('Leap_Second.dat')
         return Timescale(delta_t_recent, leap_dates, leap_offsets)
 
+    def list_bsp(self):
+        """Allow users to figure out what .bsp files are available.
+    
+            This method queries the FTP/HTTP locations given in 
+            self.urls['.bsp'] above and reutrns the available files
+            and their touch dates on the remote server.
+    
+        """
+
+        for url in self.urls['.bsp']:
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
I must finally conclude — after, alas, leaving this PR open for more than two years, which is far too long for it to have gone without a response — that this does not, to me, seem like a good fit for Skyfield. It aims to know as little as possible about outside data sources, and focuses on doing the right things with the numbers once they are downloaded.

Knowledge of how to parse external lists of catalogs for browsing might have a wide audience in Python, of course, despite not being a great fit for Skyfield itself. This code could have a future in a separate Python tool for exploring astronomy data sources. Does AstroPy have, or need, such a tool as well? It’s a generic problem that could be helpfully solved independent of either Skyfield’s or AstroPy’s own quirky ways of deciding to set up astronomy computations in Python.

Skyfield releases are sometimes already necessary because of outside data changes, so I want to avoid adding more code to Skyfield that is sensitive to how outside resources are formatted or named. Let’s as much as possible keep Skyfield a computation engine.

Thanks for the PR, though! I hope this code finds a useful home, or serves as an example for folks wanting to do the same thing from their own scripts. And thanks for your willingness to contribute, even if in this case it didn’t wind up being a perfect match for this project’s scope.

Issue #106, opened 2016-08-18
Add support for ecliptic_latlon() at additional epochs beyond J2000


hjy1210:
As far as I know, spring equinox of year 2016 is '2016-3-20 4:30'.

I check it in [HORIZONS Web](http://ssd.jpl.nasa.gov/horizons.cgi#top), get result as follows

```
Date__(UT)__HR:MN        ObsEcLon    ObsEcLat
**********************************************
$$SOE
 2016-Mar-20 04:29     359.9991687  -0.0001637
 2016-Mar-20 04:30     359.9998585  -0.0001637
 2016-Mar-20 04:31       0.0005482  -0.0001637
$$EOE
*******************************************************************************
```

But when I execute following scripts

```
# Skyfield issue
from skyfield.api import load
ts = load.timescale()
t = ts.utc(2016,3,20,4,30,0)
planets = load('de421.bsp')
sun=planets['sun']
earth.at(t).observe(sun).apparent().ecliptic_latlon()[1].degrees
```

result is:

```
359.77397918133533
```

In summary, using SKyfield, I got longitude of sun at '2016-3-20 4:30' is 359.7740, not close enough to 360.

brandon-rhodes:
The method `ecliptic_latlon()` produces coordinates relative to the J2000 ecliptic — which is why you are not finding that they line up with the equinox 16 years later. In 257d5ed I have modified the documentation to make this clear — thanks for pointing out this potential for confusion that is created by the ecliptic methods!

barrycarter:
Thought: de421.bsp is actually fairly old and may generate some inaccuracy as well.

Also, IS there a method to get ecliptic_latlon of an object based on Jdate, where date is arbitrary?

brandon-rhodes:
(a) Happily, all of the DE ephemerides are of extremely high precision for bodies like the Earth and Moon, for which we have had laser-ranging data since the Apollo landings. The difference between a 2009 ephemeris like DE421 and the more recent DE430, for example, is less than a milliarcsecond:

```
from skyfield.api import load
ts = load.timescale()
t = ts.utc(2016,3,20,4,30,0)

planets = load('de421.bsp')
sun=planets['sun']
earth=planets['earth']
lat, lon, distance = earth.at(t).observe(sun).apparent().ecliptic_latlon()
print(lon.dstr(4))

planets = load('de430t.bsp')
sun=planets['sun']
earth=planets['earth']
[...]

PR #141, opened 2017-09-22
#106 support ecliptic coordinates for the epoch of date

GammaSagittarii:
This is basically an implementation of what novas equ2ecl function does, and we get results as expected when equ2ecl function is called for the true equator and equinox of date.

I have not provided the ability to ask for coordinates for non epoch of date as ecliptic coordinates are by their nature very much tied to the dynamical system, I hope that is not a problem. 

However, we still do not match jpl horizons, because we do not match them for apparent coordinates for the epoch of date, (for example in ra, dec form) and as far as I can tell that is because they use iau76 constants, as is theorized in this discussion [https://github.com/oorb/oorb/issues/16](url) . I think because of this, matching horizons deserves its own issue.


barrycarter:
Note that the CSPICE libraries (https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/index.html) are open source and have frame transformation libraries which subsume the equ2ecl() function. Since skyfield isn't tied to ephem, might be worth looking into these. Of course, that would make Skyfield "tied to" SPICE, but I think that's OK, since SPICE generates HORIZONS.

GammaSagittarii:
I see that I have horribly underestimated the complexity of skyfield, and that one of the tests has failed in a spectacular way, because my function works only for observer on earth...

Now I know, why this wasn't implemented yet.

GammaSagittarii:
@barrycarter thank you for suggesting CSPICE, I will have its code in mind in the future.

I have fixed the issue, with tests by adding epoch choice, and I also enabled Travis for my fork so that I do not submit untested pull requests in the future :-D . The tests have passed on my fork, Travis just failed to fetch ephemeris for some 2 out of 6 builds on the main repository.

GammaSagittarii:
While my initial pull request was premature, I think that everything is in order now.
The code respects epoch parameter, it works for array of time objects and so on.
However I don't doubt that it can be further improved, and that it is quite obvious that it was written by a C programmer.

brandon-rhodes:
I am having a busy month, but am going to try to take a look at this within the next few weeks! I'll probably have more velocity once my next Python conference is over.

brandon-rhodes:
(Finally returning to a regular GitHub schedule after a few months of other commitments!)

Thanks very much for putting together working code for this improvement, @GammaSagittarii!

There are several places where this could should be tweaked before it lands — but I hesitate to offer a full code review so many months after you created this pull request, without checking first whether you're still interested or have moved on to other projects after my long delay. Are you still interested in polishing this up, in which case I'll happily go ahead and offer a few nudges in a code review? Or have I waited far to long for you to want to finish it up now, and I should take on the burden of fixing things up myself?

GammaSagittarii:
I am not really good at python so I was actually expecting that the code needs some changes. I don't mind doing the changes myself, but the thing is I cannot promise I will be able to implement them properly. I think it is better if I try to do it if that means you have more time for other stuff. (for example https://github.com/brandon-rhodes/python-jplephem/issues/14)
I kind of have an implementation of the improved precession of the ecliptic (https://www.aanda.org/articles/aa/pdf/2011/10/aa17274-11.pdf)
but I didn't send you a pull request for that, because the code is messy, if I remember correctly it doesn't work for array of dates, and I felt that you are already burdened with a lot of pull requests.

brandon-rhodes:
I'll make some quick comments, then, and you can let me know if you feel like they're in scope for what you can tackle!

skyfield/nutationlib.py
```diff
@@ -29,6 +29,32 @@ def compute_nutation(t):
...
+    If the Julian date is scalar, a simple ``(3, 3)`` matrix is
+    returned; if the date is an array of length ``n``, then an array of
+    matrices is returned with dimensions ``(3, 3, n)``.
+
+    """
+    oblm, oblt, eqeq, psi, eps = earth_tilt(t)
+    e = oblt*DEG2RAD
+    coe = cos(e)
+    sie = sin(e)
+    if not t.shape:
```
brandon-rhodes: Can you use the routine `rot_x()` in the `functions` module for this, by any chance? I tried to design that routine so that it could stand in for this entire `if` statement, and let code authors avoid having to re-implement basic rotation matrices.

skyfield/nutationlib.py
```diff
@@ -29,6 +29,32 @@ def compute_nutation(t):
...
 
+def compute_ECLIP(t):
+    """Generate the ecliptic rotations for Time `t`.
+
+    If the Julian date is scalar, a simple ``(3, 3)`` matrix is
+    returned; if the date is an array of length ``n``, then an array of
+    matrices is returned with dimensions ``(3, 3, n)``.
+
+    """
+    oblm, oblt, eqeq, psi, eps = earth_tilt(t)
```
brandon-rhodes: Oh, wow! I hadn't been thinking about ecliptic calculations when I adapted `earth_tilt()` to Skyfield and so hadn't realized that it provided the exact angle needed for the ecliptic transform.

How did you choose between `oblm` and `oblt`?

skyfield/positionlib.py
```diff
@@ -152,18 +151,61 @@ def separation_from(self, another_icrf):
...
-        d, lat, lon = to_polar(vector)
-        return (Angle(radians=lat, signed=True),
-                Angle(radians=lon),
-                Distance(au=d))
+        If you instead want the coordinates referenced to the dynamical
+        system defined by the Earth's true equator and equinox, provide
+        an epoch time.
+        """
+        lat, lon, d = self.ecliptic_latlon(epoch)
+        vector = from_polar(d.au, lat.radians, lon.radians)
```
brandon-rhodes: It feels a little bit roundabout to have `ecliptic_latlon()` (a) compute rectangular coordinates, (b) convert them to polar, and then (c) in this routine convert them back to rectangular. Could you move the `einsum` code that computes the rectangular coordinates up into this routine instead, and then have the polar routine call this one?

Also, can you eliminate the `compute_ECLIP()` function and simply call `earth_tilt()` and `rot_x()` right here?

skyfield/positionlib.py
```diff
@@ -152,18 +151,61 @@ def separation_from(self, another_icrf):
         c = dots(u1, u2)
         return Angle(radians=arccos(clip(c, -1.0, 1.0)))
 
-    def ecliptic_position(self):
-        """Compute J2000 ecliptic coordinates (x, y, z)"""
-        vector = _ECLIPJ2000.dot(self.position.au)
-        return Distance(vector)
+    def ecliptic_xyz(self, epoch=None):
```
brandon-rhodes: I think this name makes more sense than the old one! But to help Skyfield stay compatible, could you say `ecliptic_position = ecliptic_xyz` after defining the function so the old name keeps working? Thanks!

skyfield/positionlib.py
```diff
@@ -152,18 +151,61 @@ def separation_from(self, another_icrf):
...
+                epoch = self.t
+            else:
+                raise ValueError('the epoch= must be a Time object,'
+                                 ' a floating point Terrestrial Time (TT),'
+                                 ' or the string "date" for epoch-of-date')
+            position_au = einsum('ij...,j...->i...', epoch.M, position_au)
+            if not epoch.shape:
+                vector = epoch.E.dot(position_au)
+                d, lat, lon = to_polar(vector)
+                return (Angle(radians=lat, signed=True),
```
brandon-rhodes: Since every `if` ends with this same routine to build the `Angle` the same way, it feels like instead of repeating the code 3 times that you could put it down at the end of the routine outside of the `if`'s. Does that look possible to you as well?

skyfield/positionlib.py
```diff
@@ -152,18 +151,61 @@ def separation_from(self, another_icrf):
...
+                result_array = empty((epoch.E.T.shape[0], epoch.E.T.shape[1]))
+                for a in range(0, epoch.E.T.shape[0]):
+                    vector = epoch.E.T[a, 0:].dot(position_au.T[a, 0:])
+                    result_array[a, 0:] = vector
+                d, lat, lon = to_polar(result_array.T)
+                return (Angle(radians=lat, signed=True),
+                        Angle(radians=lon),
+                        Distance(au=d))
+        else:
+            vector = _ECLIPJ2000.dot(position_au)
```
brandon-rhodes: Nice! I like how you kept the routine compatible with its old behavior if `epoch` stays None.

I wonder if we could eliminate `_ECLIPJ2000` from the code, if the `earth_tilt()` routine for J2000 returns exactly the same angle as `_ECLIPJ2000` is built with?

skyfield/nutationlib.py
GammaSagittarii: I chose the one that gives the same result as novas.

skyfield/positionlib.py
GammaSagittarii: This should pose no problem.

skyfield/nutationlib.py
GammaSagittarii: Ok, it makes sense.

skyfield/positionlib.py
GammaSagittarii: I think I will be able to do it.

skyfield/positionlib.py
GammaSagittarii: Ok.

skyfield/positionlib.py
GammaSagittarii: I will try to figure it out, yes it would be nice to clean it up.

GammaSagittarii:
It seems like I will be able to do everything, and at the same time improve my skills. Thank you!

brandon-rhodes:
Thank *you*!

skyfield/nutationlib.py
GammaSagittarii: Unfortunately I ended up having problems with this part. 
If I am doing calculations with an array of times, then the angle for obliquity of the ecliptic from e = oblt * DEG2RAD is array right? So when calling rot_x(e) - I end up with an error ValueError: setting an array element with a sequence. (I am clueless as to how that everything is an array numpy magic works)

skyfield/positionlib.py
GammaSagittarii: I eliminated compute_ECLIP, however when i reorder stuff so that ecliptic_latlon calls ecliptic_xyz and converts vector to polar, I end up with this:
File "main.py", line 52, in <module>
    lat, lon, d = apparent.ecliptic_latlon(epoch='date')
File "python-skyfield/skyfield/positionlib.py", line 256, in ecliptic_latlon
    d, lat, lon = to_polar(vector)
File "python-skyfield/skyfield/functions.py", line 36, in to_polar
    r = length_of(xyz)
File "python-skyfield/skyfield/functions.py", line 24, in length_of
    return sqrt((xyz * xyz).sum(axis=0))
TypeError: unsupported operand type(s) for *: 'Distance' and 'Distance'

skyfield/positionlib.py
GammaSagittarii: Ok so I experimented with it. If I call earth_tilit for t.tdb(2000,1,1,12,0,0) which is exactly J2000,
and then I form the transformation matrix, I get this
[[1.0,  0.0,                                   0.0],
 [0.    0.9174932688587917,  0.3977513062188596],
 [0.   -0.3977513062188596,  0.9174932688587917]]
compared to what we already have:
[[1.0,  0.0,                                      0.0],
 [0.0,  0.91748206206918181, 0.39777715593191371],
 [0.0, -0.39777715593191371, 0.91748206206918181]]

I am using mean obliquity, because for some crazy reason it seems to be closer to the matrix that we already use. I can eliminate _ECLIPJ2000 if you are OK with using these values. I am not sure what do we gain by introducing unnecessary calculations though compared to fixed value that is prepared.

GammaSagittarii:
Hi, I am sorry I waited so long to work on necessary corrections.

skyfield/positionlib.py
brandon-rhodes: My guess is that you don't want the Distance object, but the `.au` attribute inside of it, which is the actual xyz position as a NumPy array. So maybe `to_polar(vector.au)`?

brandon-rhodes:
No problem! I've just responded to your comment.

GammaSagittarii:
OK I think it is done now! I figured out how to change rot_x to work with an array, like rot_z. So the only thing is left for you to decide wether you want _ECLIPJ2000 removed.

brandon-rhodes:
No, let's keep `_ECLIPJ2000`, I like including it as an option.

I see you keep sending in diffs, so let me know when you're "really done" so that I don't miss a final update :)

GammaSagittarii:
I think that is it, all tests are passing.

brandon-rhodes:
Thanks again! Before merging, I'll need at least two tests that exercise the new code paths you've created, one that does a date-driven ecliptic conversion for a single position, and one that tries a small 2-element time array.

Could you try adding a pair of quick tests like that next to `test_ecliptic_frame(ts)` in `test_against_horizons.py`? You could ask HORIZONS for the same numbers and then see how closely you match.

GammaSagittarii:
OK, so it seems I have managed to add tests. :D

brandon-rhodes:
I finally had several hours to devote to Skyfield pull requests, so this is merged — congratulations! :)

PR #146, opened 2017-12-07
Fix 2 typos and an omission in the docs

JoshPaterson:
* Add radians to list of Angle attributes
* Change two uses of `compute()` to `observe()`


brandon-rhodes:
Thanks for catching these! :)

PR #149, opened 2017-12-24
Implement __contains__ for SpiceKernel object

JoshPaterson:
Currently in order to test if an ephemeris contains a segment requires:

```
['earth'.upper()] in ephemeris.names().values()
```

Implementing this magic method allows:

```
'earth' in ephemeris
```

This commit also allows testing with integers:

```
399 in ephemeris
```


brandon-rhodes:
Thanks for this idea, it's a nice feature for a class that already has `__getitem__()`! Could you make the code a bit more efficient — instead of taking the time to turn every single code into an all-caps string, could you instead use the `_targets` to turn the string into a code and then search for it in `self.codes`? Thanks!

JoshPaterson:
Thanks! That's definitely way better!

brandon-rhodes:
So — as I get close to the next release of Skyfield, I'm wondering if this method answers exactly the right question for developers. The fact that an object _x_ is listed in an ephemeris does not, strictly speaking, mean that Skyfield will be able to generate Solar System positions for _x_ — because what an ephemeris really contains is not objects, but segments linking objects together. What an ephemeris really contains is not, say, Ganymede, but a Solar System barycenter → Jupiter barycenter segment, and then a Jupiter barycenter → Ganymede segment.

It's possible that for a given ephemeris, `__contains__()` might return `True` because the object is present as the endpoint of one segment, but `__getitem__()` would fail because there aren't the right segments present to trace the object's position all the way back to the Solar System barycenter.

So I'm wondering if I should temporarily disable the method until we figure out how it can give more accurate answers in the case of ephemerides that might not include enough segments to link all of their objects back to the Solar System barycenter. Let me know if you have any thoughts!

barrycarter:
How about `__contains__` takes two parameters, and returns whether there is a (not necessarily direct) connection between them?

brandon-rhodes:
Ah, that's a language design question, because this is a dunder method. While we could create another method that takes 2 arguments, that Python Data model dictates only a single argument for `__contains__()`:

https://docs.python.org/3/reference/datamodel.html#object.__contains__

JoshPaterson:
I think it's fine to disable this until we figure out the best way forward.

I agree that as long as Skyfield can't get a segment out of a file, then `__contains__()` should return `False` for that target name. As far as I know all of the SPK files whose data type is currently supported can tie everything back to the solar system barycenter. Here are some SPK files where this isn't true:

https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/lagrange_point/

Skyfield currently doesn't support the data type for these files, but let's say this changed and we could load these. They each contain a segment whose center is the earth-moon barycenter or the sun. The way Skyfield currently works, you could load these files, but you couldn't get any segments from them without first combining them with another file like DE430.bsp.

Do you think it would be better to change the behavior of `__getitem__()` so that if it can't link a segment back to the Solar System Barycenter, it still imports the segment and leaves it up to the user to add it to a segment from another file? Doing this would make the problem with `__contains__` moot and seems better than the current situation, which is that it's possible for there to be SPK files you can load but not get any segments from.

PR #151, opened 2018-01-20
Implement len for SpiceKernel, Angle, Distance, and Velocity

JoshPaterson:
Makes length of a SpiceKernel object the number of segments it contains. Also implements length on Angle, Distance, and Velocity objects.

### DIFF
--- a/TODO.rst
+++ b/TODO.rst
@@ -17,10 +17,6 @@ Promises for next version
 Sprint Possibilities
 ====================
 
-* Can this help us write a function to compute sub-lat/long?
-
-  https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/reclat_c.html
-
 * If we are going to allow times like TT to be submitted using
   calendar dates, then we should probably provide methods that would
   fetch them back as calendar dates.
@@ -57,10 +53,6 @@ Sprint Possibilities
   ephemeris for a raw position in au, without having to spin up a body
   object and have it spin up a `Distance` object.
 
-* A `Body` should go ahead and try building its segment list at the
-  moment it is created, instead of doing it over again every time its
-  `at()` method is called.
-
 * We currently download most SPICE kernels from NAIF, but have to use
   FTP for fetching DE422.  Are the files from the two sites equivalent
   and do they have the same data?  Should we prefer one or the other?
@@ -69,14 +61,7 @@ Sprint Possibilities
   surface, plotted against a blue atmosphere fading out into the black
   of space as the plot goes upwards towards the top.
 
-* Ephemeris objects should have more interesting repr's, that maybe give
-  the name of the ephemeris.  And maybe the objects inside?  Maybe the
-  start and end date?
-
 * Iterating across an ephemeris should probably give you object names.
-
-* The repr() of a Body object should name the body and the filename of
-  the ephemeris it is pulled from.
 
 * See about switching ``requirements.sh`` to be a conda environment
   file.
@@ -122,8 +107,6 @@ Sprint Possibilities
 * We should implement comets and asteroids using the standard formulae
   (can we find a good vector version, that will match the rest of our
   approach?) for a Keplerian orbit.
-
-* We should add formulae for the moons of the other planets.
 
 * There should be routines for downloading current astronomical data.
   Each routine should take an optional filename but should also have a
@@ -196,10 +179,7 @@ Longer-term goals
   https://github.com/brandon-rhodes/pyephem/issues/15
 
 
-.. testing
-     we need tests that handle both use_earth True and False.
-       Similarly for other variables.
-   documentation
+.. documentation
      writing up SkyField solutions to PyEphem questions on Stack Overflow
      section on accuracy of each algorithm involved
      logo?
@@ -213,5 +193,3 @@ Longer-term goals
        Could Star() become a whole catalog of stars processed in parallel?
      What routines are taking the most time when the tests are run?
      Try to take advantage of jplephem's ability to use bundles
-
-   Whether SGP4 passes the original library's test suite. [huh?]
--- a/skyfield/jpllib.py
+++ b/skyfield/jpllib.py
@@ -172,6 +172,9 @@ class SpiceKernel(object):
         else:
             code = _targets.get(name_or_code.upper())
         return code in self.codes
+    
+    def __len__(self):
+        return len(self.codes)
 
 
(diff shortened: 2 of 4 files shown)


brandon-rhodes:
At first glance, I am wary to implement `__len__()` on containers which don't support iteration, since in general in Python having a `len()` strongly implies to Python programmers that they can also use a `for` loop to iterate across your values. Could you outline your situation that led you to want to try `len(...)` on these instead of first getting the unit you wanted (`.au`, `.tt`, etc) and then taking the `len()` of that? Thanks!

JoshPaterson:
Can the other units be made iterable as well, like the Time object? I think it would be nice if the Angle, Distance, and Velocity object were consistent with the Time object in being indexable and sliceable.

There isn't anything this would allow me to do that I can't already, but sometimes I have a bunch of lines doing similar things to Time, Distance, Velocity, and/ or Angle objects, and I always have to remember that the time objects are indexable, but the others aren't. I guess it's more of an aesthetic tweak, but if it's going to mislead people than it's not worth it.

Thanks!

brandon-rhodes:
That's a good argument — there shouldn't be asymmetries between the various types. I've assigned myself to this issue, and we'll see if I or another contributor write the code to make it happen!

JoshPaterson:
The Travis tests timed out while loading jup310.bsp, similar to what happened with #152. Is there any way to increase how long Travis waits before timing out?

Anyway, this commit makes the Angle, Distance, and Velocity objects iterable and indexable. It implements __getitem__ as closely as possible to the way __getitem__ is implemented for Time objects. It also implements __iter__, but I'm not sure if it's more efficient in this case to use __iter__ or to let python default to iterating with __getitem__.  If it is in fact more efficient, should __iter__ be added for Time objects? It also includes indexing, slicing, and __len__ tests for Distance, Velocity, and Angle objects.

JoshPaterson:
I amended that last commit, triggering the tests to rerun. This time they didn't time out.

brandon-rhodes:
(Finally returning to a regular GitHub schedule after a few months of other commitments!)

Thanks for the code that works out how this would operate! Looking over the effect on the API, I am worried that it makes Skyfield less consistent, and might make it more likely for users to make errors. Let me illustrate two problems as examples.

First, let's say that I build an object that represents 4 times, and then use that to build 4 distances. My guess is that a Skyfield user would expect iteration across both the time object and the distance object would yield 4 items. But that is not quite what your code does:
```
from skyfield.api import load

ts = load.timescale()
t = ts.tt(2018, 3, (1, 2, 3, 4))

# "How many times are inside `t`?"
n = 0
for item in t:
    n += 1
print(n)

e = load('de421.bsp')
earth, mars = e['earth'], e['mars']
distance = earth.at(t).observe(mars).position

# "How many positions are inside `distance`?"
n = 0
for item in distance:
    n += 1
print(n)
```
The result of running this code would be:
```
4
3
```
I think this would surprise most users: if they've used 4 times to build 4 planetary positions, why are they only getting 3 items out when they iterate? Since users will be surprised, I don't think we should move forward with making distances iterable.

A second problem is that this approach removes a very important guard rail that was built into Skyfield. When a user wants the *x*, *y*, and *z* components of a coordinate, they have to specify the units in which they want them. This was a big design decision, because of all the problems that PyEphem users had with surprising units being returned. In Skyfield, I never wanted to read user code like this:
```
x, y, z = distance
```
because then readers would have to guess what the units were. I instead wanted them to have to write:
```
x, y, z = distance.km
```
or `au` or whatever. So I made sure that the first form of code would raise an exception that would remind them of the right thing to do.

But if we make `distance` itself iterable, we break that API decision and risk erasing important unit information from user code.

Finally, iterating in every one of these cases is very expensive: given arrays of, say, a thousand data points, it would encourage users to abandon the efficient approach of always doing array operations, and instead make it too easy to choose to spin up a thousand separate objects. I'm not sure I want to be encouraging users to take an approach whose cost a naive Python programmer won't understand, and then have to field answers to their questions about why Skyfield is so slow.

So for the above reasons, my guess is that we shouldn't make distances iterable. I wonder if we should, in the documentation, start working on a "why the API works this way" section of `skyfield/documentation/dev.rst` so that we have an explanation to point people to when they're curious why distances work the way they do, if the exceptions built into its interface don't make the reasoning clear?

I suspect that velocities and angles also shouldn't be iterable, given the above API properties, but I'm willing to hear arguments either way.

Thanks again for the code, but my guess is that we should keep Skyfield's current behavior here but better document it instead of relying on the code itself?

JoshPaterson:
Thanks for explaining this. I now understand much better why the units are the way they are, and I agree that they shouldn't be iterable or define `__len__`. Feel free to close this pull request.

I think I may have a solution that avoids these problems and still makes the units easier to use. How do you feel about adding `size` and `shape` attributes to `Time`, `Angle`, `Distance`, and `Velocity` objects that mirror the `size` and `shape` attributes of the arrays they contain? (`Time` objects already do this with `shape`.)

Here is the benefit: if I have a variable that could contain different unit objects, I could do this:

```
assert x.size == n
```

rather than this, which I sometimes have to do:

```
if isinstance(x, Angle):
    assert x.radians.size == n
if isinstance(x, Distance):
    assert x.au.size == n
```

Also, if I have a group of objects of different type, and I want to check that they are all the same length, it would allow this:

```
for item in time, distance, angle, velocity:
    assert item.size == n
```

rather than this:

```
assert time.tt.size == n
assert distance.au.size == n
assert angle.radians.size == n
assert velocity.au_per_d.size == n
```

Let me know what you think, and thanks again for taking the time to explain everything.

brandon-rhodes:
I think I like the idea! Let me think for a day, though, about whether adding `.size` and `.shape` might make them too NumPy-array-like and confuse users into thinking they can use these unit objects in other situations that arrays can be used. Then I'll plan to circle back (or by the weekend if I don't get to it tomorrow!)

Could you outline why you're using `assert` with array lengths? That's not a use case I'd heard of yet from any Skyfield users, and I'm interested in how it helps make code better, as it might give me ideas for even better supporting the pattern in the future!

JoshPaterson:
I'm getting along just fine without this, so I'm going to close it for now.

brandon-rhodes:
Thanks, hopefully we can stave off any further complexity or extra methods this way.

I must admit that I do occasionally try running `len()` on angles, and so far it hasn't been too horrible to toss in `.degrees` to fix it; if it ever gets too terrible, though, I'll know which issue to re-open!

PR #152, opened 2018-01-21
Allow Empty Time and Angle Objects

JoshPaterson:
Prior to this, empty Distance and Velocity objects were possible, but empty Time and Angle objects were not.

First, this pull request fixes the `__repr__` method in the Time object and the `dstr` and `hstr` methods of the Angle object so they still work if `self.tt` and `self.radians` are empty arrays. This was the only problem stopping empty Times and Angles from being possible before.

A second small issue it fixes was that it was possible to instantiate an Angle by providing the `radians=` keyword with a Python list, which would cause an error later when `self.radians` was expected to be a numpy array. It fixes this by making the `__init__` method of the Angle class pass arguments to the `_to_array` function, as is done in the Distance and Velocity objects.

My interest in having empty objects is that I'm working on a module for finding astronomical phenomena, like sunrise/set, twilights, stationary points, conjunctions, etc. If the desired phenomenon doesn't occur in the time range provided by the user, I've been returning `None` objects, which causes some problems. Being able to return an empty Time object would make things much cleaner. I still need to document this and clean up the tests a bit, but then I will submit it!

### DIFF
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -218,11 +218,13 @@ class Time(object):
         return self.shape[0]
 
     def __repr__(self):
-
-        if getattr(self.tt, 'shape', ()):
+        if len(self.tt)==0:
+            return '<Time []>'
+        elif getattr(self.tt, 'shape', ()):
             rstr = '<Time {0} values from tt={1:.6f} to tt={2:.6f}>'
             return rstr.format(self.tt.size, self.tt.min(), self.tt.max())
-        return '<Time tt={0:.6f}>'.format(self.tt)
+        else:
+            return '<Time tt={0:.6f}>'.format(self.tt)
 
     def __getitem__(self, index):
         # TODO: also copy cached matrices?
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -210,7 +210,7 @@ class Time(object):
     eps_correction = 0.0
 
     def __init__(self, ts, tt):
-        self.tt = _to_array(tt)
+        self.tt = tt
         self.ts = ts
         self.shape = getattr(tt, 'shape', ())
 
@@ -218,11 +218,12 @@ class Time(object):
         return self.shape[0]
 
     def __repr__(self):
-        if self.tt.size == 0:
-            return '<Time 0 values>'
-        elif getattr(self.tt, 'shape', ()):
+        size = getattr(self.tt, 'size', -1)
+        if size > 1:
             rstr = '<Time {0} values from tt={1:.6f} to tt={2:.6f}>'
             return rstr.format(self.tt.size, self.tt.min(), self.tt.max())
+        elif size == 0:
+            return '<Time []>'
         else:
             return '<Time tt={0:.6f}>'.format(self.tt)
 
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -218,8 +218,8 @@ class Time(object):
         return self.shape[0]
 
     def __repr__(self):
-        if getattr(self.tt, 'shape', ()) and self.tt.shape == (0,):
-            return '<Time []>'
+        if self.tt.size == 0:
+            return '<Time 0 values>'
         elif getattr(self.tt, 'shape', ()):
             rstr = '<Time {0} values from tt={1:.6f} to tt={2:.6f}>'
             return rstr.format(self.tt.size, self.tt.min(), self.tt.max())
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -210,7 +210,7 @@ class Time(object):
     eps_correction = 0.0
 
     def __init__(self, ts, tt):
-        self.tt = _to_array(tt)
+        self.tt = tt
         self.ts = ts
         self.shape = getattr(tt, 'shape', ())
 
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -218,7 +218,7 @@ class Time(object):
         return self.shape[0]
 
     def __repr__(self):
-        if len(self.tt)==0:
+        if getattr(self.tt, 'shape', ()) and self.tt.shape == (0,):
             return '<Time []>'
         elif getattr(self.tt, 'shape', ()):
             rstr = '<Time {0} values from tt={1:.6f} to tt={2:.6f}>'
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -210,7 +210,7 @@ class Time(object):
     eps_correction = 0.0
 
     def __init__(self, ts, tt):
-        self.tt = tt
+        self.tt = _to_array(tt)
         self.ts = ts
         self.shape = getattr(tt, 'shape', ())
 
--- a/skyfield/units.py
+++ b/skyfield/units.py
@@ -12,6 +12,8 @@ def _to_array(value):
     """As a convenience, turn Python lists and tuples into NumPy arrays."""
     if isinstance(value, (tuple, list)):
         return array(value)
+    elif isinstance(value, (float, int)):
+        return np.float64(value)
     else:
         return value
 
@@ -188,8 +190,8 @@ class Angle(object):
         raise AttributeError('no attribute named %r' % (name,))
 
     def __str__(self):
-        if getattr(self.radians, 'shape', ()) and self.radians.shape == (0,):
-            return '[] radians'
+        if self.radians.size == 0:
+            return '0 values'
         return self.dstr() if self.preference == 'degrees' else self.hstr()
 
     def __repr__(self):
@@ -221,7 +223,7 @@ class Angle(object):
         """Convert to a string like ``12h 07m 30.00s``."""
         if warn and self.preference != 'hours':
             raise WrongUnitError('hstr')
-        if getattr(self.radians, 'shape', ()) and self.radians.shape == (0,):
+        if self.radians.size == 0:
             return '0 values'
         hours = self._hours
         if getattr(hours, 'shape', ()):
@@ -258,7 +260,7 @@ class Angle(object):
         """Convert to a string like ``181deg 52\' 30.0"``."""
         if warn and self.preference != 'degrees':
             raise WrongUnitError('dstr')
-        if getattr(self.radians, 'shape', ()) and self.radians.shape == (0,):
+        if self.radians.size == 0:
             return '0 values'
         degrees = self._degrees
         signed = self.signed
--- a/skyfield/units.py
+++ b/skyfield/units.py
@@ -191,11 +191,14 @@ class Angle(object):
 
     def __str__(self):
         if self.radians.size == 0:
-            return '0 values'
+            return 'Angle []'
         return self.dstr() if self.preference == 'degrees' else self.hstr()
 
     def __repr__(self):
-        return '<{0} {1}>'.format(type(self).__name__, self)
+        if self.radians.size == 0:
+            return '<{0} []>'.format(type(self).__name__, self)
+        else:
+            return '<{0} {1}>'.format(type(self).__name__, self)
 
     def hms(self, warn=True):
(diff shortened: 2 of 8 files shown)


JoshPaterson:
I have to fix the way I run tests, and then I will fix this.

skyfield/timelib.py
```diff
@@ -218,11 +218,13 @@ def __len__(self):
         return self.shape[0]
 
     def __repr__(self):
-
-        if getattr(self.tt, 'shape', ()):
+        if getattr(self.tt, 'shape', ()) and self.tt.shape == (0,):
```
brandon-rhodes: 1. In the usual case that the array is non-empty, this new code performs 3 lookups of `.shape` — could you start out with a statement `shape = …` to run `getattr` only once, then re-use the value each time you need it?

2. Would it make sense to test `self.tt.size == 0` instead, so that empty n-dimensional arrays get the `[]`, and not just 1-dimensional empty arrays?

(And so forth with the other instances of this pattern.)

JoshPaterson: I like the idea of testing for the size attribute, which would work great for Time objects. However, if Angle, Distance, and Velocity objects are initialized with a python float, then self.radians, self.au, and self. au_per_d become python floats, which don't have the size attribute. This is the situation I was trying to avoid by testing if the `shape` attribute exists before calling it. For these objects, can floats be converted to numpy float64's before being used to initialize the object? Perhaps by adding another case to the `_to_array` function? If that's possible I think that's the best solution.

JoshPaterson:
Even with the changes I just made, it's still possible to initialize a Time object with a float like this:
`Time(ts, 143525.25)`
and then self.tt would not have the size attribute, leading to an error in the __repr__ method. I wasn't sure how best to address this. Maybe by converting the float to a float64, but where is the best place to put the conversion?

One other consideration is that empty angle objects would still have a preference for degrees vs. hours. I can't think of problems that would cause, can you?

skyfield/timelib.py
brandon-rhodes: Alas, my explanation was too short! I was only trying to suggest that the second term of your `and` be adjusted — you are quite right that you need the initial check for shape, protected by `getattr`. I was thinking of:
```
        shape = getattr(self.tt, 'shape', ())
        if shape and self.tt.size == 0:
```
That would protect against the case of a plain float, I think?

brandon-rhodes:
> One other consideration is that empty angle objects would still have a preference for degrees vs. hours. I can't think of problems that would cause, can you?

Could you outline the steps of how that would happen? The current test `'hours' if hours is not None` should, I think, still select `hours` if that's the array that was provided, since even an empty array will not test positive for `None`?

JoshPaterson:
Sorry, I phrased that very confusingly. I should have said that "empty angle objects would still have a preference for either degrees or hours". What I'm trying to get at is that we currently have two types of empty angles, some that prefer to use dstr, and some that prefer to use hstr, despite the fact that hstr and dstr return the same result for an empty Angle. It feels to me like an empty Angle should be degree/hour agnostic, but then I also can't think of any problems caused by empty Angles having a preference. Just wanted to run that by you to see if you had any thoughts.

brandon-rhodes:
Thanks for the clarification! Since NumPy arrays are not immutable, I imagine that the array behind an angle could later gain elements even if it was empty to begin with. In that case, the preference the angle object has for degrees versus hours would be useful, because it would know how to present the angle values that had been appended. So I don't see a problem with an empty angle having a preference already — does that make sense?

JoshPaterson:
That makes sense, I can see how that would be useful!

I just added one more commit that fixes the issue I brought up a few comments ago about when a Time object is initiated with a python float. That clears up all the issues with this pull request that I know about.

Can you try rerunning the Travis tests? I don't have the button to do so since I don't have write access. It appears Travis failed because it timed out while loading jup310.bsp. It passes all the tests when I run them locally.

brandon-rhodes:
I am going to try to clean up the testing story this weekend. We shouldn't really need to download a huge file each time. I wonder if I could put little excerpts from each ephemeris directly in the repository, and use those for the tests?

Then I could have you rebase atop the tests and everything would go green. I'll let you know if I make progress!

skyfield/timelib.py
```diff
@@ -210,19 +210,21 @@ class Time(object):
...
         self.ts = ts
         self.shape = getattr(tt, 'shape', ())
 
     def __len__(self):
         return self.shape[0]
 
     def __repr__(self):
-
-        if getattr(self.tt, 'shape', ()):
+        if self.tt.size == 0:
```
brandon-rhodes: I'm not seeing how this avoids raising an exception immediately if `self.tt` is a normal Python float?

skyfield/timelib.py
```diff
@@ -210,19 +210,21 @@ class Time(object):
     eps_correction = 0.0
 
     def __init__(self, ts, tt):
-        self.tt = tt
+        self.tt = _to_array(tt)
```
brandon-rhodes: Could you outline the use case you ran into that made this necessary? I'm not sure I've seen a Skyfield user ever instantiate `Time()` object directly — the library's design has them create `Time` objects through the methods of a `Timescale`, which I think has `_to_array()` everywhere it's needed. If one of them is missing, I'd prefer to fix it there rather than add a second call here.

skyfield/timelib.py
```diff
@@ -210,19 +210,21 @@ class Time(object):
...
         self.shape = getattr(tt, 'shape', ())
 
     def __len__(self):
         return self.shape[0]
 
     def __repr__(self):
-
-        if getattr(self.tt, 'shape', ()):
+        if self.tt.size == 0:
+            return '<Time 0 values>'
```
brandon-rhodes: Nice, I'm glad that this will no longer cause an error — thanks! I'm still sometimes tweaking the repr's and str's, so I might adjust this more later, but for now it should do fine.

skyfield/units.py
```diff
@@ -142,7 +144,8 @@ def to(self, unit):
 Angle(degrees=value)
 Angle(hours=value)
 
-where `value` can be either a Python float or a NumPy array of floats"""
+where `value` can be either a Python float, a list of Python floats,
```
brandon-rhodes: Thanks for updating the docs!

skyfield/units.py
```diff
@@ -187,6 +190,8 @@ def __getattr__(self, name):
         raise AttributeError('no attribute named %r' % (name,))
 
     def __str__(self):
+        if self.radians.size == 0:
+            return '0 values'
```
brandon-rhodes: Having a zero `0` in the string might make folks confused, since it looks like a quantity, and `0 values` won't communicate to them what kind of object they've printed out. How could we make the output string include the type, so that, like the normal str and repr, this communicates what kind of value this is? `Empty Angle` or something like that?

skyfield/vectorlib.py
```diff
@@ -69,6 +69,9 @@ def at(self, t):
             raise ValueError('please provide the at() method with a Time'
                              ' instance as its argument, instead of the'
                              ' value {0!r}'.format(t))
+        if getattr(t.tt, 'shape', ()) and t.tt.shape == (0,):
+            raise ValueError('the time object provided to the at() method is '
```
brandon-rhodes: Could you outline why this should raise an error? Thanks! (I ask because there are lots of values that could wind up empty in Skyfield, but I'm not sure how a normal user would come up with an empty `Time` object to begin with, and if we add this check then we'll have to face the question of whether every Skyfield call everywhere needs a check for empty inputs, and that's a path I'm wary of because it's several dozen more lines of code for an error I'm not sure yet how real world users would run into.)

skyfield/timelib.py
JoshPaterson: I did this so that `self.tt` could never be a python float, even if someone instantiates a Time object directly. If users never do that, then this wouldn't be necessary.

skyfield/timelib.py
JoshPaterson: This doesn't avoid raising the exception, instead it depends on `self.tt` never being a python float, which is why I changed line 213 to use `_to_array`. If users never instantiate a Time object directly, then `self.tt` being a python float is very unlikely, and maybe this is fine as is? Otherwise it should change to be able to handle that.

skyfield/units.py
JoshPaterson: My intent here was to match the existing way all of the `str`'s and `repr`'s work. I didn't notice that this is the only case that doesn't include units implying the object type. I think something like `<Angle [empty]>`, or `<Angle []>` would be fine here.

skyfield/vectorlib.py
JoshPaterson: I was trying to think of any problems empty objects could cause, and I guess I was being a little overzealous. It's not necessary, so if you have doubts leave it out.

skyfield/units.py
brandon-rhodes: I like `<Angle []>`, great idea! Try pivoting to that pattern, I think it will look good and communicate clearly.

skyfield/vectorlib.py
brandon-rhodes: Understood! Let me know if you'd like to remove it when you update the pull request to change the repr's we discussed above, or if I should try pulling it out as I merge.

skyfield/timelib.py
brandon-rhodes: Understood. Try removing the `_to_array()` up above, and then pivoting this to be `getattr(self.tt, 'size', 1)` so that this class remains, for the moment at least, agnostic about whether it's been given a Python float or not. Thanks!

brandon-rhodes:
Thanks for your responses! I've been busy adding a big feature to the underlying jplephem library, which has increased my ping time a bit up here at the Skyfield level of the stack. I've made a few final comments that should vector the pull request in to where I'll then land it. Thanks!

JoshPaterson:
I fixed the Time object, and I tried to have the `repr` use as few function calls as possible, but I'm not sure if that's what you had in mind. I also made the `str` for empty Angle objects `Angle []` instead of `<Angle []>` because the strings for non-empty angles don't have brackets. If you want to change any of this before pulling, that's totally fine with me. Thanks!

JoshPaterson:
`AssertionError: '[1.23 4.56] au' != '[ 1.23  4.56] au'`

This error happened only in the tests for Python 3.5, and yet passed in some of the other python versions. I'm not sure why.

brandon-rhodes:
Yes, I am seeing those assertion errors too, also only for Python 3.5! I'll let you know if I get time to dig in and figure out what produces them.

JoshPaterson:
Is there any chance this will make it into the next release?

brandon-rhodes:
Yes, there's a chance — because I think I've fixed the tests now! Please rebase and let's see if we can get you green and ready to land.

brandon-rhodes:
Yes! I'll go ahead and merge it now — thanks for the prod.

brandon-rhodes:
(Wrote two earlier comments too quickly, then re-read the pull request and now am answering for real!)

Yes, there's still a chance, if you could address the comments that are still outstanding: that I'd like the Time init method to not yet call `_to_array()` because it's a bit of delay that I'm not sure the official API ever needs; and I don't think it should be an error to create objects for empty arrays.

Thanks for the work on this! Also, the build is now green, if you'd like to rebase and see the CI go green too. :)

JoshPaterson:
Sorry about that, I didn't mean to leave those comments unaddressed. I've read over the conversation again, and as far as I can tell I've fixed everything.

Thanks!

JoshPaterson:
The 3.5 check seems to have timed out while attempting to load jup310.bsp.

brandon-rhodes:
Thanks so much for being willing to do several round trips on this! I'll see about getting a Skyfield release out soon now that we have it merged.

JoshPaterson:
Awesome! Thanks!

Issue #125, opened 2017-04-01
Add "epoch of position" parameter for stars


pkgw:
Sometimes the positional information that you have for a celestial object is anchored to an epoch other than J2000.0. This matters if the object has a non-zero proper motion or parallax.

As far as I can tell, Skyfield doesn't support "epochs of position" other than J2000.0. It would be nice to add this — I think it's just a matter of wiring in some parameters in `starlib.py` that would replace `T0`. I can prepare a PR for this if there's interest.

brandon-rhodes:
Good point! We should broaden Skyfield’s support for other epochs. Right now, the only code that adapts coordinates in one epoch to the J2000.0 epoch that Skyfield uses internally is an obscure function in a work-in-progress module that adapts Hipparcos records, whose proper motion is referred to J1991.25, to J2000.0.

The Skyfield routine, like much of Skyfield, is based on the USNO’s work in their NOVAS library. Search for the `transform_hip` routine’s description in their documentation:

http://aa.usno.navy.mil/software/novas/novas_c/NOVAS_C3.1_Guide.pdf

and then in their C code:

http://aa.usno.navy.mil/software/novas/novas_c/novas.c

If you want all the details. The Skyfield routine that does the same thing lives in:

https://github.com/skyfielders/python-skyfield/blob/master/skyfield/data/hipparcos.py

The code should certainly be generalized to handle input coordinates whose proper motion is from other epochs besides the exact epoch used by Hipparcos. In addition, I suppose people might want a convenient way to import catalogs whose coordinate epoch is not J2000.0?

Do you have specific catalogs of your own that you were wanting to import, that we could target specifically?

pkgw:
I'm running into this when computing present-day apparent positions for sources in the 2MASS catalog, since their astrometry is referenced to the date when they were observed. The 2MASS source catalog comes with a `Date` observation that gives that information.

I guess I was envisioning that the `Star` class would acquire a new property, something like `epoch_of_position`, that would default to J2000.0. Then the uses of `T0` in [Star._observe_from_bcrs](https://github.com/skyfielders/python-skyfield/blob/master/skyfield/starlib.py#L79) would just use `self.epoch_of_position` instead. I have an intuitive feeling that this would be a bit nicer way to do things than by a one-time operation that "corrects" records à la the Hipparcos code, since it preserves more information. But that's certainly not a deeply-considered stance.

brandon-rhodes:
[...]

PR #166, opened 2018-03-28
Allow Stars to have epochs of position other than T0.

pkgw:
Pursuant to issue #125. I have documented the new keyword but not added test coverage.

### DIFF
--- a/skyfield/documentation/stars.rst
+++ b/skyfield/documentation/stars.rst
@@ -107,12 +107,15 @@ Here are all three effects applied to Barnard’s star:
 
 .. testcode::
 
+    from skyfield.api import T0
+
     barnard = Star(ra_hours=(17, 57, 48.49803),
                    dec_degrees=(4, 41, 36.2072),
                    ra_mas_per_year=-798.71,
                    dec_mas_per_year=+10337.77,
                    parallax_mas=545.4,
-                   radial_km_per_s=-110.6)
+                   radial_km_per_s=-110.6,
+                   jd_of_position=T0)
 
     astrometric = earth.at(t).observe(barnard)
     ra, dec, distance = astrometric.radec()
@@ -126,10 +129,12 @@ Here are all three effects applied to Barnard’s star:
     +04deg 44' 01.3"
     3.77863e+05 au
 
-Note that the above position is different than the input right ascension
-and declination but not because we have asked for dynamic coordinates.
-This position is in fixed ICRS coordinates and indicates real motion on
-the part of Barnard’s star across our sky.
+(The ``jd_of_position`` parameter defaults to ``T0`` but it is used
+above to call attention to its existence.) Note that the above position
+is different than the input right ascension and declination but not
+because we have asked for dynamic coordinates. This position is in fixed
+ICRS coordinates and indicates real motion on the part of Barnard’s star
+across our sky.
 
 See the guide to :doc:`positions` to learn the operations that you can
 perform with these astrometric positions after using a :class:`Star`
--- a/skyfield/starlib.py
+++ b/skyfield/starlib.py
@@ -40,7 +40,7 @@ class Star(object):
 
     def __init__(self, ra=None, dec=None, ra_hours=None, dec_degrees=None,
                  ra_mas_per_year=0.0, dec_mas_per_year=0.0,
-                 parallax_mas=0.0, radial_km_per_s=0.0, names=()):
+                 parallax_mas=0.0, radial_km_per_s=0.0, jd_of_position=T0, names=()):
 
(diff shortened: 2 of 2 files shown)


skyfield/starlib.py
```diff
@@ -40,7 +40,7 @@ class Star(object):
 
     def __init__(self, ra=None, dec=None, ra_hours=None, dec_degrees=None,
                  ra_mas_per_year=0.0, dec_mas_per_year=0.0,
-                 parallax_mas=0.0, radial_km_per_s=0.0, names=()):
+                 parallax_mas=0.0, radial_km_per_s=0.0, jd_of_position=T0, names=()):
```
brandon-rhodes: What would you think of naming this `epoch` to agree with the terminology of the Hipparcos catalog?

skyfield/documentation/stars.rst
```diff
@@ -107,12 +107,15 @@ Here are all three effects applied to Barnard’s star:
...
+    from skyfield.api import T0
+
     barnard = Star(ra_hours=(17, 57, 48.49803),
                    dec_degrees=(4, 41, 36.2072),
                    ra_mas_per_year=-798.71,
                    dec_mas_per_year=+10337.77,
                    parallax_mas=545.4,
-                   radial_km_per_s=-110.6)
+                   radial_km_per_s=-110.6,
+                   jd_of_position=T0)
```
brandon-rhodes: I suspect we should expand this to two separate examples, one without, and one with the new keyword, since many users cut-and-paste without reading the text. But I can do that after this lands if you don't get around to it!

skyfield/starlib.py
pkgw: I chose this version because it makes clear the units of the input value, in the same way the `_km_per_s` suffixes on some of the arguments to this function do.

skyfield/documentation/stars.rst
pkgw: As per Twitter conversion from yesterday and today, I suspect? I mostly avoided this since an example that was *almost* entirely redundant seemed a bit lame, but I didn't want to go to the work of coming up with and validating another example :-)

skyfield/starlib.py
brandon-rhodes: 1. Maybe `epoch_jd`, then?
2. Drat, now I'm thinking of the much bigger question of whether this should take a `Time` object :) It doesn't need to, right? There's other places in the code already where the interface just sticks to a plain jd?

brandon-rhodes: (Just checking back in: before we land, do you have an opinion on the spelling here, and the question of allowing a Time vs just a float?)

pkgw: Thanks for the nudge. I don't like `epoch` as the root word of the name for this parameter since in my opinion it's more vague to the uninitiated — but I'm not sure how to make that choice anything more than a matter of opinion. In which case I think "package owner's opinion wins" is the only sensible resolution process ...

And, yeahhhh, if there are Time objects, it should probably take a Time object. It was sheer laziness on my part not to look into how to make that work. I'd presume it's pretty easy — just convert it to a JD and save that?

brandon-rhodes: Let's try `epoch` and I'll see what it looks like as I try using it after it lands! If they pass a float, then just keep it untouched; but if they pass a time object from Skyfield's `timelib`, I think you can ask it for its `.tt` attribute to get the float you need. Thanks!

pkgw: The `epoch` value is subtracted from a TDB time. Should I ask for the Time's `.tdb` attribute instead? (I am wondering if maybe asking for TDB opens up a can of worms.)

brandon-rhodes: Oh! Yes, then `.tdb` is the correct attribute, thanks!

skyfield/starlib.py
```diff
@@ -40,7 +41,7 @@ class Star(object):
 
     def __init__(self, ra=None, dec=None, ra_hours=None, dec_degrees=None,
                  ra_mas_per_year=0.0, dec_mas_per_year=0.0,
-                 parallax_mas=0.0, radial_km_per_s=0.0, names=()):
+                 parallax_mas=0.0, radial_km_per_s=0.0, epoch=T0, names=()):
```
brandon-rhodes: Oh, there's one final adjustment needed — I should have noticed before: `epoch` needs to go last, else any code that folks have written that supplies plain non-keyword arguments for a `Star` will break if we take the argument that was once `names` and use that same slot for `epoch` instead. Try moving `epoch=` to the end so that existing calls keep working. Thanks!

pkgw: Done.

Issue #171, opened 2018-04-12
Backup filename check for deltat does not take into account path


joernu76:
When supplying the Loader class with a directory on Windows 7, the renaming of deltat.preds fails:

  File "C:\Users\icg173\Anaconda2\envs\python3\lib\site-packages\skyfield\iokit.py", line 159, in __call__
    os.rename(self.path_to(filename), self.path_to(backup_name))
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\icg173\\.config\\mss\\deltat.preds' -> 'C:\\Users\\icg173\\.config\\mss\\deltat.old1.preds'

The problem seems to be a missing "self.path_to" here:
skyfield/iokit.py, line 156ff
[...]
                    if not os.path.exists(backup_name):
                        break
                self._log('  Renaming to: {0}', backup_name)
os.rename(self.path_to(filename), self.path_to(backup_name))
[...]
should be changed to:
[...]
                    if not os.path.exists(self.path_to(backup_name)):
[...]

I also end up with many backups after fixing it, but that seems to be related to issue #170

JoshPaterson:
I also get this error on Windows 10.

PR #172, opened 2018-04-12
Fix bug in checking for existing backup path of deltat file.

joernu76:
Fix issue #171


brandon-rhodes:
The tests don't look terribly happy, but this makes so much sense as I read it that I'm going to go ahead and land and clean up any consequences later. Thanks!

Issue #163, opened 2018-03-22
Unable to load TLE without name line


Engineero:
When using `load.tle()`, it would be nice to add a check to either grab the first line as the satellite name (if three lines are provided per TLE), or just parse out the satellite ID and use that as the name if the list is only TLEs. For instance, if I download the [latest TLE list from space-track.org][1] (two-line elements, not three), to `'satcat.txt'` and try to `load.tle('./satcat.txt')`, I get the following traceback:

> ---------------------------------------------------------------------------
> AssertionError                            Traceback (most recent call last)
> /usr/local/lib/python3.6/site-packages/sgp4/io.py in twoline2rv(longstr1, longstr2, whichconst, afspc_mode)
>     130     try:
> --> 131         assert line.startswith('1 ')
>     132         satrec.satnum = int(line[2:7])
> 
> AssertionError: 
> 
> During handling of the above exception, another exception occurred:
> 
> ValueError                                Traceback (most recent call last)
> <ipython-input-3-d64f2d7594bb> in <module>()
> ----> 1 tles = load.tle('./satcat.txt')
> 
> /usr/local/lib/python3.6/site-packages/skyfield/iokit.py in tle(self, url, reload)
>     195         """
>     196         with self.open(url, reload=reload) as f:
> --> 197             return dict(parse_celestrak_tle(f))
>     198 
>     199     def open(self, url, mode='rb', reload=False):
> 
> /usr/local/lib/python3.6/site-packages/skyfield/iokit.py in parse_celestrak_tle(fileobj)
>     336         line1 = next(lines).decode('ascii')
>     337         line2 = next(lines).decode('ascii')
> --> 338         sat = EarthSatellite(line1, line2, name)
>     339         yield name, sat
>     340         if ' (' in name:
> 
> /usr/local/lib/python3.6/site-packages/skyfield/sgp4lib.py in __init__(self, line1, line2, name, ts)
>      79     def __init__(self, line1, line2, name=None, ts=None):
>      80         self.name = None if name is None else name.strip()
> ---> 81         sat = twoline2rv(line1, line2, whichconst=wgs72)
>      82         self.model = sat
>      83         if ts is None:
> 
> /usr/local/lib/python3.6/site-packages/sgp4/io.py in twoline2rv(longstr1, longstr2, whichconst, afspc_mode)
>     151         # elnum = int(line[64:68])
>     152     except (AssertionError, IndexError, ValueError):
[...]

PR #177, opened 2018-05-10
Feature/load nameless tle

Engineero:
Added the ability to load a nameless TLE using `load.tle()` from the API package. For example, this allows us to load:

```
1 00005U 58002B   18130.23524677  .00000070  00000-0  10021-3 0  9993
2 00005  34.2455  77.6488 1845867 131.1918 246.3184 10.84813857121819
```

rather than:

```
0 VANGUARD 1
1 00005U 58002B   18130.23524677  .00000070  00000-0  10021-3 0  9993
2 00005  34.2455  77.6488 1845867 131.1918 246.3184 10.84813857121819
```

The resulting `EarthSatellite` object will then have the `name` field set to the satellite ID number, in this case `'00005'` without the classification letter at the end.

I also improved the check/make directory logic in the `Loader` to be more robust to possible race conditions/access from multiple sources.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -361,7 +361,7 @@ def parse_celestrak_tle(fileobj):
         else:  # two-line element set, no name provided!
             line1 = line.decode('ascii')
             line2 = next(lines).decode('ascii')
-            name = line1.split()[1][:-1]  # grab satellite ID# as name
+            name = line2.split()[1]  # grab satellite ID# as name
         sat = EarthSatellite(line1, line2, name)
         yield name, sat
         if ' (' in name:
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -188,10 +188,14 @@ class Loader(object):
     def tle(self, url, reload=False):
         """Load and parse a satellite TLE file.
 
-        Given a URL or a local path, this loads a file of three-line
-        records in the common Celestrak file format, where each first
-        line gives the name of a satellite and the following two lines
-        are the TLE orbital elements.
+        Given a URL or a local path, this loads a file of two- or three-line
+        records in the common Celestrak file format. For a three-line element
+        set, each first line gives the name of a satellite and the following
+        two lines are the TLE orbital elements. A two-line element set
+        comprises only these last two lines.
+
+        If two-line element sets are provided, the EarthSatellite 'name'
+        attribute is set to the satellite ID number for the object.
 
         Returns a Python dictionary whose keys are satellite names and
         values are :class:`~skyfield.sgp4lib.EarthSatellite` objects.
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -362,11 +362,12 @@ def parse_celestrak_tle(fileobj):
             name = line.decode('ascii').strip()
             line1 = next(lines).decode('ascii')
             line2 = next(lines).decode('ascii')
+            sat = EarthSatellite(line1, line2, name)
         else:  # two-line element set, no name provided!
             line1 = line.decode('ascii')
             line2 = next(lines).decode('ascii')
-            name = line2.split()[1]  # grab satellite ID# as name
-        sat = EarthSatellite(line1, line2, name)
+            sat = EarthSatellite(line1, line2, name=None)
+            sat.name = str(sat.model.satnum)  # set to satellite number
         yield sat.model.satnum, sat
         yield name, sat
         if ' (' in name:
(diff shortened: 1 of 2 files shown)


Engineero:
Resolves issue #163

skyfield/iokit.py
```diff
@@ -184,10 +188,14 @@ def _log(self, message, *args):
     def tle(self, url, reload=False):
         """Load and parse a satellite TLE file.
 
-        Given a URL or a local path, this loads a file of three-line
-        records in the common Celestrak file format, where each first
-        line gives the name of a satellite and the following two lines
-        are the TLE orbital elements.
+        Given a URL or a local path, this loads a file of two- or three-line
+        records in the common Celestrak file format. For a three-line element
```
brandon-rhodes: We'll want to maybe remove "Celestrak" here if we're accepting 2-line files, since Celestrak always uses 3 lines.

skyfield/iokit.py
```diff
@@ -350,9 +358,14 @@ def parse_leap_seconds(fileobj):
 def parse_celestrak_tle(fileobj):
     lines = iter(fileobj)
     for line in lines:
-        name = line.decode('ascii').strip()
-        line1 = next(lines).decode('ascii')
-        line2 = next(lines).decode('ascii')
+        if line.decode('ascii').strip()[0] == '0':
```
brandon-rhodes: I can't quite tell what the comparison to `'0'` is trying to accomplish — clicking through to one sample Celestrak file, I'm not seeing a leading `0` character on any of its lines:

https://celestrak.com/NORAD/elements/stations.txt

skyfield/iokit.py
```diff
@@ -350,9 +358,14 @@ def parse_leap_seconds(fileobj):
...
+            name = line.decode('ascii').strip()
+            line1 = next(lines).decode('ascii')
+            line2 = next(lines).decode('ascii')
+        else:  # two-line element set, no name provided!
+            line1 = line.decode('ascii')
+            line2 = next(lines).decode('ascii')
+            name = line2.split()[1]  # grab satellite ID# as name
```
brandon-rhodes: Instead of doing a `split()` here that repeats the work that `EarthSatellite` will do to parse the name, let's try pivoting to `name = None` here and then set the name after the satellite object has been created by pulling its own opinion of the satellite ID that it will have parsed out from the input lines, so that we don't do the same work twice. Thanks!

brandon-rhodes:
I've just gone in and committed a basic test for the TLE reading logic. You'll now have a pattern you can follow for adding a second test that tries out the same routine but against this alternative format you're helping Skyfield support. Good luck, hopefully it will be straightforward!

skyfield/iokit.py
Engineero: I will reword for my next commit.

skyfield/iokit.py
Engineero: Sorry, I'll look at the celestrak format specifically. The source that I was using for my elsets (space-track.org) always starts the name line with a "0", then a "1" for the first TLE line and a "2" for the second. I was under the impression that that was the standard.

skyfield/iokit.py
Engineero: Will do! I just have to figure out the right way to access and set these.

Engineero:
Thanks for reviewing! And I should have included tests in my pull request, sorry about that. I will address your comments here and add some tests. I may make a couple of intermediate commits, so I will try to tag you in a note in the discussion when I think it is ready to go again.

Engineero:
Sorry, still a bit of a noob with some of this collaborative stuff. How do I run your tests?

Engineero:
@brandon-rhodes updates all submitted and tests are passing!

brandon-rhodes:
Thanks very much for this contribution! I'll go ahead and merge, even though I'm still thinking through whether we should backfill a name for satellites that don't have one; and if we do, should it really be just a bare integer, or maybe say "ID ..." with the integer. Let me know in a comment here the advantages of the approach you pioneered, if you weighed it against other options, and we'll see if I wind up adjusting it or not.

Thanks again for the contribution; I didn't have access to this file format!

Engineero:
I don't really have a strong argument for using the satellite ID as the name, other than that it seemed obvious to me as a choice for name when an actually satellite name string was not available. We might be able to build a utility to query space-track.org for the names of satellites for which we only have ID and provide that as a separate function call.

I have been building a spacetrack API for doing this sort of thing that we could use, or we could just build something custom. Either way, I would make it a separate call to populate your EarthSatellites with names since querying space-track.org requires authentication information.

brandon-rhodes:
Thanks for the further thoughts! For the moment I'm leaving name `None` if there wasn't a name in the file, since if we fill it in from the satellite number then there won't be an absolutely reliable way to tell, when looking at the satellite, whether the file provided a name (that happened to be the number) or didn't provide one at all. If it really starts tripping users up in an unforeseen way we can revisit, but for this first try let's go with completely unambiguous, and see where that takes us.

Thanks again!

PR #185, opened 2018-06-29
Complete subpoint elevation_m calculation.

alz:
Calculates satellite altitude above subpoint.

From Dr. T.S. Kelso's explanation (https://www.celestrak.com/columns/v02n03/)


dsholes:
This looks correct. A quick comparison between ISS TLE propagation using skyfield with this updated subpoint() method (including elevation_m calculation) versus http://www.isstracker.com/ shows good agreement.

brandon-rhodes:
Thanks!

PR #186, opened 2018-06-29
Fix for Lat/Lon calculation #182

### LINKED ISSUE #182: Lat/Lon from subpoint
In the function to get lat/lon of a subpoint, the `reverse_terra` function is called. I have looked at the code and think that this could lead wrong values in case the `arctan2` function returns a negative value (for lon).
`lon = (arctan2(y, x) - 15 * DEG2RAD * gast) % tau` 
Replacing above line with:
`lon = (((arctan2(y, x) - 15 * DEG2RAD * gast) + tau / 2.0) % tau) - tau / 2.0`
should fix the issue. Output will then be `+/-pi`

Am I wrong here?

see https://github.com/dieli/python-skyfield/pull/1

dieli:
This should fix #182

### DIFF
--- a/skyfield/charting.py
+++ b/skyfield/charting.py
@@ -3,8 +3,10 @@
 #import numpy as np
 from .starlib import Star
 
-def plot_stars(catalog, observer, project, mag1, mag2, ax):
+def _plot_stars(catalog, observer, project, ax, mag1, mag2, margin=1.25):
+    """Experiment in progress, hence the underscore; expect changes."""
 
+    art = []
 
     # from astropy import wcs
     # w = wcs.WCS(naxis=2)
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thank you! So that I can write a test of the new behavior to keep it fixed as the code continues to evolve, could you share (here in a comment would be fine; I can add the test) a particular circumstance for which the current routine returns the wrong longitude value?

alz:
`from skyfield.api import EarthSatellite, load

text = """
ISS (ZARYA)
1 25544U 98067A   18180.89340144  .00002199  00000-0  40642-4 0  9992
2 25544  51.6425 315.3706 0003869 243.2106 259.9289 15.53990257120490
"""
lines = text.strip().splitlines()
ts = load.timescale()
iss = EarthSatellite(lines[1], lines[2], lines[0])
geo = iss.at(ts.utc(2018, 6, 30, 20, 45, 0))
print("{}".format(geo.subpoint().longitude.degrees))`

Output:

260.38333100901275

PR #192, opened 2018-07-30
Support for CIRS coordinates

### LINKED ISSUE #104: Enhancement: support for CIRS coordinates
While I'm on an issue submitting roll, a suggested enhancement....

For the telescope I work on we're attempting to make the switch to using Skyfield (which is awesome, thanks!) for our time/ephemeris calculations. At the same time it makes most sense for us to use coordinates in the Celestial Intermediate Reference System (CIRS), that is using the Celestial Intermediate Origin (CIO) as the RA origin, and not the vernal Equinox. The CIRS is the alternative (and probably more sensible) approach in all the new IAU resolutions. I think it would be great if Skyfield had support for generating coordinates in this dynamical system.

My suggestions for how to go about doing this (at least as a start)...

Add a `cirs` option to `skyfield.positionlib.ICRF.radec`, e.g.

``` python
ICRF.radec(self, epoch=None, cirs=False)
```

An alternative could be an `origin` argument

``` python
ICRF.radec(self, epoch=None, origin='equinox')
```

which can also take the value `'cio'`. If this is turned on (by either method), and an epoch is given, the RA/DEC returned are CIRS coordinates in that epoch instead.

This also requires that the transformation from CIRS to GCRS (and vice-versa) is added. In USNO Circular 179, this is a matrix called **C**, as opposed to the GCRS to equinox conversion **M**. I would suggest adding in attributes for `C` and `CT` to `Time` in same way as `M` and `MT` are included.

The calculation of `C` probably requires a few additional routines to be added, e.g. a routine to calculate the _equation of origins_ (which can probably be constructed out of the routines in `precession lib.py`, or one for calculating the _CIO locator, s_.

Anyway, if you'd be happy to accept a Pull Request for this, I should be able to spend some time putting it together.

Thanks again!

jrs65:
Here's a patch for CIRS coordinates adding a `cirs` option to `.radec`, and adding the `C` and `CT` matrices to `Time` objects. There's also a set of tests. Some are internal, one is testing against an external calculation using SOFA (this agrees at the 10 micro-arc-second level.

Please let me know if there's any more changes or suggestions that you have.

Fixes #104


skyfield/positionlib.py
```diff
@@ -91,7 +91,7 @@ def speed(self):
         """
         return Velocity(length_of(self.velocity.au_per_d))
 
-    def radec(self, epoch=None):
+    def radec(self, epoch=None, cirs=False):
```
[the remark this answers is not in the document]
jrs65: Sounds good. Although I have another suggestion I'll put in another comment.

jrs65:
Your comment about using a `.cirs_ra()` method got me thinking a little bit more about the semantics of how to do this.

I think an alternative approach which is maybe more consistent with the idea of the equinox-based and CIRS representing frames in their own right would be to add them as specific frame classes. Something like this approximate sketch:
```python
class EquinoxSystem(ICRF):
    ...

class CIRS(ICRF):
    ...
```

And then add two methods to ICRF that transform the instance into the new frame:
```python

def equinox(self, time):
    # Multiply by time.M to get pos etc...
    return EquinoxSystem(pos)  

def cirs(self, time):
    # Multiply by time.C to get pos etc...
    return CIRS(pos)
```
and modify the `radec` method to be like:

```python
def radec(self, epoch=None):

    # For backwards compatibility essentially does what it always did
    if epoch is not None:
        return self.equinox(epoch).radec()

    position_au = self.position.au

    r_au, dec, ra = to_polar(position_au)
    return (Angle(radians=ra, preference='hours'),
                 Angle(radians=dec, signed=True),
                 Distance(r_au))
```

Overall this would allow making calls like:
```python
# CIRS ...
ra, dec, d = earth.at(st).observe(star).cirs(st).radec()

# Equinox ...
ra, dec, d = earth.at(st).observe(star).equinox(st).radec()
# ... equivalent to the existing ...
ra, dec, d = earth.at(st).observe(star).radec(st)
```

It's certainly a little bit more complex, but I think it's a little more conceptually clear, but easy to make backwards compatible such that if a user doesn't care about the distinction it just does what they'd always thought it should do.

brandon-rhodes:
That's a great question, and I should add developer docs addressing this. The thinking behind the design is this:

With previous libraries, I found that new users get confused very quickly about when two different sets of numbers represent truly different positions in the sky (like the astrometric vs the apparent position for a planet; truly two different locations amongst the stars) and when different numbers are simply different ways of naming a single position.

So Skyfield's answer is to never create two different Python objects for the same position. Instead, any given place-in-the-sky gets a single position object and only ever a single position object, with as many methods as we want for turning those into numbers.

So, by this scheme, CORS coordinates must be merely one more of the many methods on the single position object that, whatever the coordinates used to represent it, is exactly and precisely "one place in the sky".

Thanks for asking, it's a very important design principle that I had been needing to write up! Maybe this weekend I can paste this explanation into a file. :)

brandon-rhodes:
All right! A talk I gave a couple of weeks ago at PyBay gave me practice describing the Skyfield approach that avoids creating separate objects for separate coordinates, which I've now made a part of the official Design document:

https://github.com/skyfielders/python-skyfield/commit/b4e3b514a3ee2a2805fbf056c84ee0444eb4d47e

Now that I've better surfaced that rationale, I'm going to officially put this pull request in "Request changes" mode so that it drops temporarily off of my GitHub dashboard. As soon as you're ready to make the quick pivot to a separate method, simply request review and I think we'll be ready to merge.

Thanks for making this new functionality available in Skyfield!

skyfield/positionlib.py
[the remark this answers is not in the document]
brandon-rhodes: Thanks the the comment and for the suggestion that there's a different way forward here (making objects for coordinates, not just methods). But since I suspect that many Skyfield users are well served by the distinction between positions (objects) and coordinates (methods), let's move forward with the plan to make this a separate method. Since some folks might want the xyz coordinates, feel free to break this into two methods `cirs_xyz()` and `cirs_radec()` like ecliptic and galactic coordinates do.

I look forward to merging the result — thanks again!

brandon-rhodes:
You're right, the error has nothing to do with your diff! Hopefully I'll have a few minutes to review it in the morning :)

skyfield/positionlib.py
```diff
@@ -157,6 +157,43 @@ def separation_from(self, another_icrf):
...
 
+    def cirs_xyz(self, epoch):
+        """Compute cartesian CIRS coordinates at a given epoch (x, y, z).
+
+        Calculate coordinates in the Celestial Intermediate Reference System
+        (CIRS), a dynamical coordinate system referenced to the Celestial
+        Intermediate Origin (CIO). As this is a dynamical system it must be
+        calculated at a specific epoch.
+        """
+        if epoch is not None:
```
brandon-rhodes: What if the opposite is true — what if `epoch` is indeed `None`?

brandon-rhodes: (Oh — and, I have a vague feeling I should really have named this parameter `equinox` this whole time. What do you think? Does that sound more correct? Should we go ahead and name it that here?

jrs65: Good point, should have thought about this more carefully. I've made it throw the same `ValueError` if it gets `None`.

As for the parameter name ,`epoch` is probably the correct terminology.

jrs65:
One thing I've noticed just now: the definitions in `to_polar` are unconventional and don't match the documentation. I'll create a separate issue for it.

skyfield/positionlib.py
brandon-rhodes: My worry is this wikipedia article (which could of course be half incorrect):

"Equinox is often confused with epoch, with the difference between the two being that the equinox addresses changes in the coordinate system, while the epoch addresses changes in the position of the celestial body itself"

https://en.wikipedia.org/wiki/Equinox_(celestial_coordinates)

But it's okay for now to simply make this match the rest of the API.

brandon-rhodes:
Merged! Thanks for this contribution :)

skyfield/positionlib.py
jrs65: Hmmmm. I've never come across this distinction before. That nomenclature might make sense for traditional equinox based coordinates, though probably maintaining consistency for the moment is more important.

For CIO based coordinates (e.g. the CIRS) using `equinox` might be very confusing as it's a coordinate system designed to be separated from the actual equinoxes!

PR #194, opened 2018-08-01
Fix bug in `Loader.open` method

uhho:
`os.path.remove(path)` is throwing an error.
Should be `os.remove(path)`.


brandon-rhodes:
Thanks! As I was writing this code, I was planning how I'll write some tests for it as a next goal for Skyfield, and obviously it needed them.

Issue #11, opened 2013-08-14
Add ICRS position calculation for bodies in keplerian orbits


hoylemd:
repo:
https://github.com/hoylemd/python-skyfield/tree/keplerian_orbits

paper on the task:
http://ssd.jpl.nasa.gov/txt/aprx_pos_planets.pdf

brandon-rhodes:
Happily, this feature was just released in a new version of Skyfield:

https://rhodesmill.org/skyfield/kepler-orbits.html

hoylemd:
Haha what a coincidence. I was just thinking about Skyfield and @brandon-rhodes randomly today! Congrats on the release :)

brandon-rhodes:
Thanks! I hope the intervening years have been good to you, and that you have enjoyed the niches you’ve occupied in software since then. I’ll be happy to buy you a beer the next time we’re at a conference together!

hoylemd:
Thanks! I know I'd enjoy catching up :)

PR #202, opened 2018-09-15
Fix Docs for Orbital Elements

JoshPaterson:
``longitude_of_ascending_node`` was missing and ``longitude_of_periapsis`` was listed twice.

I also rearranged the categories such that using one element from each of the 6 categories fully defines an orbit. Added a note mentioning this.

### DIFF
--- a/skyfield/documentation/elements.rst
+++ b/skyfield/documentation/elements.rst
@@ -47,13 +47,17 @@ Here is a list of the attributes of the Elements object and their types:
 
 .. parsed-literal::
 
-    Elements describing orientation of the orbit plane:
-     ├── ``inclination``                 → Angle object
-     └── ``longitude_of_periapsis``      → Angle object
+    Element describing shape of the orbit:
+     └── ``eccentricity``                → numpy.ndarray
 
-    Elements describing shape and direction the orbit within its plane:
+    Element describing the tilt of the orbital plane:
+     └── ``inclination``                 → Angle object
+
+    Element describing the direction in which the orbital plane is tilted:
+     └── ``longitude_of_ascending node`` → Angle object
+
+    Elements describing direction of periapsis:
      ├── ``argument_of_periapsis``       → Angle object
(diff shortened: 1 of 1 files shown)


JoshPaterson:
Also, because the inline code font size is smaller than the rest of the text, the arrows in the table don't align vertically. Do you know the best way to fix this?

brandon-rhodes:
What happens if you just take the backtick markers off of the names — would that solve the formatting issue? Check out `positions.rst` for a different pattern that should rescue the vertical alignment — and thanks for asking; that's a nice detail to try to get right on the rendered site!

Since you're back looking at orbital elements again: I think Skyfield might next, if you're interested, benefit from logic that could propagate an orbit to predict a future position. Looking at the SPICE source code, I was interested to see that the state-of-the-art is, it appears, to skip trying to use trigonometry and traditional orbital elements, and to use a more robust technique involving pure vector operations (if I understood the code on its first read-through).

It might be the most profusely commented astronomy code I've seen yet:

http://localhost:8003/spicelib/prop2b.html

https://github.com/magnific0/cspice/blob/master/src/cspice/prop2b.c

Let me know if you'd be interested in adapting this code to Python; it would finally let us close #11, which might be the oldest still-open issue here in the Skyfield project! :)

brandon-rhodes:
Yes, I think they could be! At least judging from the comments, the author(s?) seem to have thought very thoroughly through the mathematics and even to have improved upon the maths of the book they are using as a source. As I think (?) that SPICE is the library behind the JPL's HORIZONS system, which seems to be the gold standard for generating ephemerides online, adopting their math would not only let us benefit from the work they've done on improving the steps, but would also give us easy agreement with the numbers people run on HORIZONS.

I'm tempted, once we have a pure vectors-to-vectors propagation routine working, to jettison the old trig-based approach in `skyfield/contrib/keplerianlib.py` which the contributor could never get to agree with HORIZONS.

PR #203, opened 2018-09-15
Fix reference_frames keyword in elementslib

JoshPaterson:
Fixes the reference_frame keyword of `osculating_elements_of`, which currently isn't working. This also makes the rest of the tests (currently marked broken) pass.


brandon-rhodes:
Thanks for the fix! :)

PR #204, opened 2018-09-16
Make OsculatingElements attribute table look better

JoshPaterson:
I changed the attributes to regular text instead of inline code so the arrows would line up. I also made the lines continuous between categories of elements so the table would be more similar to the table in ``positions.rst``.


brandon-rhodes:
Thanks!

PR #211, opened 2018-10-14
Another approach to finding astronomical phenomena

JoshPaterson:
This is my approach to finding astronomical phenomena. It's vastly improved over the earlier versions I submitted to ``astronomy-notebooks`` (mostly in terms of speed). If ``skyfield.almanac`` can benefit by taking anything from it that would be great. If not, that's ok too. My intention is for this PR to be a place for discussion; I don't so much care whether this ultimately gets pulled into the contrib folder.

I think the advantages of this approach are that:

* the api is simpler
* the same functions work for ``VectorSum`` objects, ``EarthSatellite`` objects, and ``Star`` objects
* The search algorithms are faster

I've included an example script (almanac2_examples.py) to demonstrate the api.

Based on my limited understanding of phalanx search, I believe this approach in effect performs one round of phalanx search in order to narrow in on regions of the objective function that contain the points of interest, and then submits those regions to the vectorized secant and brent_min functions.

Based on comparing the two versions of the ``seasons`` and ``moon_phases`` functions, the vectorized secant method is about 40% faster than the phalanx method, while producing sub millisecond accuracy.

``risings_settings`` is slower in this approach than the ``skyfield.almanac`` equivalent, because it first calculates all upper and lower culminations in the time range, and then uses those times as partitions in order to ensure that if the object just peeks over the horizon it won't be missed. I'm still working on understanding how ``skyfield.almanac`` can be sure not to miss any rises and sets without doing this.


skyfield/contrib/almanac2/almanac2_examples.py
```diff
@@ -0,0 +1,48 @@
+from skyfield.api import Loader, Topos, EarthSatellite, Star
+from almanac2 import seasons, moon_phases, meridian_transits, culminations, twilights, risings_settings
+
+load = Loader(r'C:\Users\Josh\Scripts\Skyfield_Data')
```
gyth: maybe a more explicit line 4 would be welcome ?

JoshPaterson: Thanks for pointing this out, I should have mentioned it before. The tests and the example script require de430t.bsp, which is quite large (130 MB). I didn't want that to automatically start downloading to anyone's skyfield/contrib/almanac2 directory, so in order to run ``test_almanac2.py`` or ``almanac2_examples.py`` you'll have to edit them so that the Loader instance uses whichever directory you keep your .bsp files. I don't know of a way to avoid automatically downloading de430t.bsp and still allow those scripts to run without needing modification first. I just changed this so that there's an empty string instead of the path from my computer.

You should still be able to import from ``almanac2.py`` without any modifications.

JoshPaterson:
The tests shouldn't be failing, since I didn't change anything except add new files. It looks like there might be a problem with pip failing in the python 3.3 build.

barrycarter:
I googled "phalanx search" and found nothing useful. Could you provide a link? Just as a note, CSPICE has a wide array of search functions you might want to look at-- if yours is faster, it would be quite impressive!

JoshPaterson:
Phalanx search is the method used in ``skyfield.almanac``. Here's a talk that Brandon gave describing it along with ``skyfield.almanac`` in general: 

https://www.youtube.com/watch?v=cYHP3oRGjG8

The search functions used in this approach are vectorized versions of the the [secant root finding function](https://github.com/brandon-rhodes/pyephem/blob/f96daf12d4f815be92e0caa52611b444517b9e0d/ephem/__init__.py#L91) from PyEphem, and [brent's method for finding minimums](https://github.com/scipy/scipy/blob/c23542ac7b4b9660f1f2adb6dcccb4fe451e3824/scipy/optimize/optimize.py#L1896) from ``scipy.optimize``.

The first version of these functions from last year that I put in skyfielders/astronomy-notebooks used the root and minimum finding functions from scipy, but it was extremely slow. This version is about 15 times faster than that due to two main changes: 

The first problem is that skyfield is vectorized and produces data much more efficiently if you ask it for data for many times at once instead of one time, while the scipy optimization functions are not vectorized and can only use scalars as arguments to the objective function. Brandon discusses this in the talk I linked to above. A few weeks ago I finally got vectorized versions of these functions working, and that change sped up the total time to run the tests by about 7x.

The second problem was that in order to find the moon phases for example, the secant function had to be called separately for each different value of the difference in ecliptic longitude (0, 90, 180, 270 degrees). To allow all of these times to be found in one call of the secant function and therefore in one round of iterations I added the [function ``g``](https://github.com/skyfielders/python-skyfield/pull/211/files#diff-1acbad2864fd3a172ffd7b6fac5f44c7R48), which is a transformation of the objective function such that all of the partitions have a root at zero instead of the value that is being searched for. The ``brent_min`` function also makes a similar transformation such that all of the partitions contain minimums and since it is vectorized, they can all be found at once, thus making the fewest possible calls to the objective function. This change sped up the tests by another 2x.

I don't know whether either of those two changes are applicable to CSPICE, I'm not familiar enough with how it works internally. Since the secant method for finding roots and Brent's method for finding minimums have been around for so long I'm assuming CSPICE already takes full advantage of them!

JoshPaterson:
Travis seems to be having trouble connecting to ssd.jpl.nasa.gov to download ephemerides, which is causing the tests to fail.

brandon-rhodes:
Yes, I think that this weekend I'll finally take the plunge and commit the necessary ephemerides to the repository. I won't distribute them with Skyfield — that would make the package much too big! — but I think it's reasonable for folks who check out the repository to expect to have the disk space for DE421.

brandon-rhodes:
The necessary files are now committed to the repository and the build is green again! See if rebasing helps your own build now. :)

JoshPaterson:
Thank you! I feel better now having that green check :)

brandon-rhodes:
(Oh, it's under `contrib` so it can't be imported as part of normal Skyfield. Nice.)

PR #212, opened 2018-10-16
consider adding north, east down and east, north, up local coordinate system outputs

anielsen001:
I was interested in adding local tangent plane coordinate systems besides atl, az, distance so I created a north, east, down and east, north, up cartesian coordinate system.

I don't know that these are the best ways to implement these options. If there is interest in options like this, I can work to improve the implementation.

### DIFF
--- a/skyfield/__init__.py
+++ b/skyfield/__init__.py
@@ -5,6 +5,6 @@ functions and classes there.  See the ``documentation`` directory inside
 the source code, as well as the http://rhodesmill.org/skyfield/ site!
 
 """
-VERSION = (1, 8)
+VERSION = (1, 9)
 __version__ = '.'.join(map(str, VERSION))
 
--- a/skyfield/positionlib.py
+++ b/skyfield/positionlib.py
@@ -1,6 +1,6 @@
 """Classes representing different kinds of astronomical position."""
 
-from numpy import arccos, array, clip, einsum, empty, exp, full, nan
+from numpy import arccos, array, clip, einsum, empty, exp, full, nan, array
 
 from .constants import ANGVEL, DAY_S, DEG2RAD, RAD2DEG, tau
 from .data.spice import inertial_frames
@@ -329,7 +329,16 @@ class Geometric(ICRF):
         return _to_altaz(self.position.au, self.observer_data,
                          temperature_C, pressure_mbar)
 
+    def east_north_up( self ):
+        """ return (east, north, up) cartesian coordinates in the local 
+        reference frame, using kilometers
+        """
+        R = self.observer_data.altaz_rotation
 
+        position_km = einsum('ij...,j...->i...', R, self.position.km)
+        position_km = array([ position_km[1], position_km[0], position_km[2] ])
+        return position_km        
+        
 class Barycentric(ICRF):
     """An (x, y, z) position measured from the Solar System barycenter.
 
--- a/skyfield/tests/test_positions.py
+++ b/skyfield/tests/test_positions.py
@@ -1,8 +1,12 @@
+from __future__ import print_function
+
 import numpy as np
 from skyfield import api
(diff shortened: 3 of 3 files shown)


barrycarter:
The ENU seems like a good idea, and mirrors what Mathematica does: https://reference.wolfram.com/language/ref/GeoPositionENU.html

The NED seems almost redundant though? East, North, and Up are three perpendicular axes and any position can be expressed using these (and negative numbers for West, South, and down)?

As a fun note: because the Earth is an ellipsoid, "Down" doesn't point to the center of the Earth (the midpoint of the foci that define it's ellipse), nor is it in the same direction that gravity pulls. CSPICE refers to this Up vector as the "surface normal".

anielsen001:
I don't have a specific preference either way, but NED has the convenience that the angle from x toward y matches the stand azimuth/heading angle people use when talking about geography.

brandon-rhodes:
I do like the idea of exposing Cartesian coordinates here, like we already do for ecliptic and galactic coordinates! Some initial thoughts:

1. It would be nice to have only one system to support instead of two, since folks can flip a sign and axis and get any other system they wish.
2. "Down" would seem to be an unusual choice; every expression I've seen of a local coordinate system uses "Up" instead.
3. We would want to add some tests.
4. I'm not sure 3-letter method names will be enough when people read code a year later, so whichever choice we make, let's think about spelling it out like `north_east_up()` or something.

Thanks for suggesting this idea!

anielsen001:
I agree that spelling out the names would make more sense to most people and that maintaining two is harder. I will look at creating some tests and make some updates to see what people think.

brandon-rhodes:
Nice, I like that method name! Could you add a quick test that, say, observes the sky's North Pole from a spot on the equator to see if the vectors turn out correct? Thanks!

anielsen001:
Yes, I'm working on a test. I will hopefully be able to commit it soon.

anielsen001:
I'm a bit flummoxed and I hope I can get some help. I'm trying to understand what's happening when I compute the difference between two Topos objects. That is, something like this: 
```
north_pole = Topos( latitude_degrees = 90.0, longitude_degrees = 0.0 )
eqtr = Topos( latitude_degrees = 0.0, longitude_degrees= 0.0) 
diff = north_pole - eqtr
```

then I compute
```
diff.at( ts.utc( datetime.now().replace(tzinfo= utc))).position.km
```
And if I do this repeatedly, the result changes like this:
```
In [298]: diff.at( ts.utc( datetime.now().replace(tzinfo= utc))).position.km
Out[298]: array([-6128.20443291,  1727.7175515 ,  6367.81148927])

In [299]: diff.at( ts.utc( datetime.now().replace(tzinfo= utc))).position.km
Out[299]: array([-6128.2752056 ,  1727.46605625,  6367.81161004])

In [300]: diff.at( ts.utc( datetime.now().replace(tzinfo= utc))).position.km
Out[300]: array([-6128.34973339,  1727.20117432,  6367.81173722])

In [301]: diff.at( ts.utc( datetime.now().replace(tzinfo= utc))).position.km
Out[301]: array([-6128.42860231,  1726.92081567,  6367.81187181])
```
I imagined that since both points are on the earth surface, the relative distance between them will remain fixed with time, but it doesn't. This is how I perform my tests, for the east_north_up so I need to understand this point.

anielsen001:
I believe that I've now got a working version with tests.

brandon-rhodes:
I'll assume that the contributor is at the moment too busy to proceed with this PR — which, on a close re-reading, would require more work to get into shape than I had originally recognized: it seems to use the position vector as “up” whereas in most cases it is not in exactly that direction, and sometimes is not even close, I don't think? I'm going to therefore close this, and if in the future someone opens an issue needing east-north-up, then we can make a fresh try at it.

Thanks for the idea, though, and someday it might indeed make it into Skyfield!

PR #236, opened 2019-02-01
Correctly parse deltat.preds file using new format

reallyasi9:
As of today, seemingly, the published deltat.preds file has changed to include the Julian date.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -338,20 +338,25 @@ def parse_deltat_data(fileobj):
 def parse_deltat_preds(fileobj):
     """Parse the United States Naval Observatory ``deltat.preds`` file.
 
-    Each line gives a floating point year, the value of Delta T, and one
-    or two other fields::
+    Each line gives a floating point Julian date, a floating point year, the
+    value of Delta T, and one or two other fields::
 
-    2015.75      67.97               0.210         0.02
+  MJD         YEAR    TT-UT Pred   UT1-UTC Pred  ERROR
+   58119.000  2018.00   69.06       0.122       0.117
+   58210.000  2018.25   69.22      -0.033       0.162
+   58301.000  2018.50   69.36      -0.177       0.215
+   58392.000  2018.75   69.45      -0.268       0.273
+   58484.000  2019.00   69.62                   0.335
 
     This function returns a 2xN array of raw Julian dates and matching
     Delta T values.
 
     """
-    year_float, delta_t = np.loadtxt(fileobj, skiprows=3, usecols=[0, 1]).T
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thank you for letting me know! I'll try to release a fixed version today.

I'm not sure I can merge this code as-is, though, because thousands of machines will already have a `deltat.preds` on disk that's in the old format — which I think this branch would break? The routine will have to become able to detect the difference between the two formats, I think.

brandon-rhodes:
Thanks again so much for notifying me quickly! I have just released Skyfield 1.10. If you could, test it for me and let me know if it now parses the file correctly. Thanks!

zorrolo:
Hi, it seems like this is relevant for issues I have with my scripts. Unfortunately none of them work anymore and as I tried to execute the simplest example from the skyfield documentation it produces the very same error on my machine.
As I am a student who recently just started to learn python and using skyfield in order to achieve some calculations for planets position and distance to earth I don't really have enough knowledge to help myself.
Could someone please explain (in a beginner friendly way) what exactly to do to avoid this error?

Here is the simplest example that leads to the mentioned error:
`#!/usr/bin/env python
# -*- coding: utf-8 -*-

from skyfield.api import load

planets = load('de431_excerpt2.bsp')
earth, mars = planets['earth'], planets['mars barycenter']

ts = load.timescale()
t = ts.now()
astrometric = earth.at(t).observe(mars)
ra, dec, distance = astrometric.radec()

#print(ra)
#print(dec)
print(distance)`

and this is the error message I receive every time:
`$ python python3-script.py 
Traceback (most recent call last):
  File "python3-script.py", line 9, in <module>
    ts = load.timescale()
  File "/home/user/anaconda3/lib/python3.7/site-packages/skyfield/iokit.py", line 270, in timescale
    preds = self('deltat.preds')
  File "/home/user/anaconda3/lib/python3.7/site-packages/skyfield/iokit.py", line 151, in __call__
    expiration_date, data = parser(f)
  File "/home/user/anaconda3/lib/python3.7/site-packages/skyfield/iokit.py", line 347, in parse_deltat_preds
    expiration_date = date(year[0] + 2, month[0], 1)
ValueError: year 58668 is out of range
`

brandon-rhodes:
@zorrolo If you'll upgrade Skyfield to the latest version, the error should go away — maybe `pip install -U skyfield` will do it?

zorrolo:
@brandon-rhodes Thank you very much! I tried to install skyfield before but what I didn't use was that '-U' parameter. Thankfully it worked when I tried it using your suggestions. Now everything seems to work perfectly fine again!

brandon-rhodes:
Good, I'm glad it's working — enjoy!

troyswanson:
Thanks for this fix!

Issue #105, opened 2016-08-16
Run offline with uniform timescale?


uhoh-SE:
In [this answer][1] I mentioned that one needs to be connected to the internet the first time to download an ephemeris (and other things), and then Skyfield will run off line - for a while.

If someone wanted to access all of Skyfield's features but was willing to give up automatic checking for leap seconds and either had some work-around, or just could live with it, is there some way - any way - that Skyfield be run long-term off-line*?

I understand that a very appropriate answer might be along the lines of "even if there were, it would be bad to do that, so *just say no!*"

It looks like it's always necessary to use `ts` but there is no way to convince it to keep working if the leap second info is out of date by more than Skyfield's internal grace period.

(*) imagine just for example a solar/battery powered Raspberry Pi that points a mirror such that moonlight passes through a window and down a hallway whenever it's possible. That's a piece of cake with Skyfield, but it would still currently require an internet connection if I understand correctly.

[1]: http://astronomy.stackexchange.com/a/17946/7982

brandon-rhodes:
Alas — I added such logic back and documented it back in July (see 267f0038f29b45bf72a8eb960151d391f49e2ca0), but I have not released a new Skyfield since then to let people start using an offline mode!

I will run through the open issues and pull requests today and try to get a release out by this evening. In the meantime, you can use the new features described in the above commit immediately by running:

    pip install https://github.com/skyfielders/python-skyfield/archive/master.zip

Let me know if the new feature works as you expect!

PR #251, opened 2019-04-17
Revert filename change if download fails

gustavmpettersson:
In previous changes (see Issue #105) support was added to run Skyfield offline by creating a Loader which does not check file expiry dates. Some use cases (like mine, which is a portable satellite tracker) has intermittent internet access (and/or servers can be down). Therefore it is desirable to enable users to attempt the download, but use old files if this fails. I.e:

```
try:
    ts = skyfield.api.Loader('some_directory').timescale()
except OSError:
    ts = skyfield.api.Loader('some_directory', expire=False).timescale()
```

However, the current version of Skyfield will rename any existing files to 'filename.old' before attempting downloading, and not revert this. This pull request simply adds reverting to the old filename to make the above usecase possible.

Specifically:
Old behaviour: When an expired file is found it is renamed to 'file.oldX' and a download attempt is made. If this fails the program exits with an OSError.
New behaviour: If said download fails the file is renamed again to its old name. The program then exits with an OSError.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -182,7 +182,18 @@ class Loader(object):
                              .format(filename))
 
         self._log('  Downloading: {0}', url)
-        download(url, path, self.verbose)
+        try:
+            download(url, path, self.verbose)
+        except OSError as download_err:
+            self._log('  Failed to download')
+            # If we made a backup revert the name change.
+            try:
+                os.rename(self.path_to(backup_name), self.path_to(filename))
+                self._log('  The backup was restored')
+            except:
+                self._log('  Backup not restored')
+                pass
+            raise download_err
 
         if parser is not None:
             self._log('  Parsing with: {0}()', parser.__name__)


brandon-rhodes:
Adding myself as a reviewer — this looks like a good idea, I'll hopefully have time to review this weekend!

gustavmpettersson:
Just a gentle bump here :) Did you find a moment to review this merge?

brandon-rhodes:
Alas, no; I probably won't have time to study it until after my keynote in London this coming weekend, which has been consuming my free time!

brandon-rhodes:
@gustavmpettersson — Looking back at this PR after the passage of time, I am thinking of a different approach to this problem: if the file already exists, then doing the download to a file that includes `.new` in its name, and only if the download succeeds doing the renames of the old file then the new one.

Feel free to reply within the next day if you would like to try pivoting this PR in that direction. Otherwise, I'll be happy to close it for you, and try making the improvement myself. I'll understand completely if you're off on other projects, as the original PR is more than a year old at this point, so feel free to not answer!

gustavmpettersson:
Hi Brandon, as guessed I'm off in other directions, but I do like your solution and will close this.

brandon-rhodes:
Thanks for replying! I will hopefully have time in July for a bit of improvement here.

brandon-rhodes:
Skyfield 1.26, released at the beginning of August, reworks the download logic to solve this using an improved approach: it simply doesn't move the old file out of the way until the download succeeds. Hopefully this will prevent the inconvenient behavior described in this issue from recurring in the future!

PR #253, opened 2019-05-19
[WIP] Propagating Kepler Orbits

JoshPaterson:
After a few false starts I finally have a vectorized function that propagates a pair of position and velocity vectors with an array of times. It is based on the function prop2b from the Spice library.

One lesson I learned is that the original Fortran is much more readable than the C version, because the C code is automatically generated from the Fortran code. For example this Fortran code:
```
         MAXC   = MAX   ( 1.0D0, DABS(BR0),
     .                           DABS(B2RV),
     .                           DABS(BQ),
     .                           DABS(QOVR0/BQ) )
```

became this in C:
```
	d__2 = 1., d__3 = abs(br0), d__2 = max(d__2,d__3), d__3 = abs(b2rv), 
		d__2 = max(d__2,d__3), d__3 = abs(bq), d__2 = max(d__2,d__3), 
		d__3 = (d__1 = qovr0 / bq, abs(d__1));
	maxc = max(d__2,d__3);
```

The reason I'm making this PR is that I'd like to discuss what uses this can be put to and what API's those uses should have. I'm thinking that there should be a `KeplerOrbit` object that would be the Skyfield equivalent of PyEphem's `EllipticalBody`, `ParabolicBody`, and `HyperbolicBody`, and also a way to create Pandas dataframes of the data from https://minorplanetcenter.net/data, and a way to create `KeplerOrbit` objects from rows of those dataframes.

I included the original Fortran function and the non-vectorized python version of `keplerlib.propagate` in the `contrib` folder because comparing the three versions might help answer questions about why `keplerlib.propagate` does things the way it does.

The tests work by starting with a set of orbital elements, converting those to state vectors, propagating the state vectors, converting back to orbital elements, and then making sure the orbital elements are still the same within some tolerance.


JoshPaterson:
I'm not quite sure what is going on with the CI, 2.6 and 2.7 are failing because of code in this PR, 3.4 passes, and 3.5 and 3.6 are failing for other reasons.

brandon-rhodes:
Wow, thank you for doing this work! Let's not commit the Fortran code; you can link to the original file from your new module's docstring. I agree that a Pandas-based dataframe would be a very nice addition, but I also agree that landing the basic mechanism first makes the most sense.

I agree that a KeplerOrbit object holding the parameters is a good approach; when its `.at()` method is called, it can return a position centered at the Solar System barycenter. Once it exists, I can iterate to make it work with `observe()` too, in a separate diff.

One test I would like added is to give the object the coordinates of an object from HORIZONS, and to test that it returns the same position that HORIZONS does. When writing tests, I generally use as the sample test date the same date that I'm writing the test :)

JoshPaterson:
No problem! I'm glad to help. Here's what I've got so far for a working `KeplerOrbit` object. I took a stab at creating a complete API, which seems to work properly.

The numbers being produced are in the ballpark of Horizons, but I think they ought to be closer. I'm currently writing tests and trying to figure out why they aren't closer.

Here's how the API works as currently written:

Imports:
```
from skyfield.api import load
from skyfield.keplerlib import KeplerOrbit

ts = load.timescale()
ephem = load('de430t.bsp')
sun = ephem['sun']
```

Create a dataframe of asteroids using any file from https://minorplanetcenter.net/data that ends in '_extended.json.gz':
```
asteroids_url = 'https://minorplanetcenter.net/Extended_Files/nea_extended.json.gz'
asteroids = load.mpcorb(asteroids_url)
```

The same location has a file containing data about comets, but it has a different format, so it has it's own load method:

```
comets_url = 'https://www.minorplanetcenter.net/Extended_Files/cometels.json.gz'
comets = load.mpc_comets(comets_url)
```

`KeplerOrbit` objects can be created from rows of these dataframes like this:
```
eros = KeplerOrbit.from_mpcorb_dataframe(asteroids.loc[0], ts)
hale_bopp = KeplerOrbit.from_comet_dataframe(comets.loc[0], ts)
```

Because the data from the MPC are heliocentric the `KeplerOrbit`s created with them are centered at the sun. They can be combined with an ephemeris segment for the sun to produce positions in the ICRS like this:
```
(sun + hale_bopp).at(ts.utc(2019, 6, 1)).position
```

brandon-rhodes:
Just a quick note to let you know that I'm reading your comments on this issue, and am interested to hear whether you are able to find the discrepancy with HORIZONS. If not then maybe we could land the code as a not-yet-turned-on feature and you could continue to iterate?

I will have to think about how this makes the I/O code more complicated, and whether we should pivot towards not needing `reload` and `filename` options repeated over and over again in I/O methods. Maybe it was a design mistake for me to couple "download" and "open" in the first place? I'll think about it; but that needn't block this landing once it agrees with HORIZONS, since I can adjust the I/O approach before the next Skyfield release.

brandon-rhodes:
> The numbers being produced are in the ballpark of Horizons, but I think they ought to be closer. I'm currently writing tests and trying to figure out why they aren't closer.

I thought should check back and see whether you were ever able to track down the discrepancies between coordinates returned by your code, and those returned by HORIZONS?

JoshPaterson:
Sorry I’ve been neglecting this for a while, but I think I’ll to be able to devote some more time to this in the next few weeks.

I was able to get the results from this module to more closely match horizons. I don’t expect them to match exactly however, because this module uses 2-body propagation, while I believe that Horizons uses numerical integration and takes into account the masses and positions of many solar system objects.

The only source of exact 2-body propagation data that I know of is the function `prop2b` from the Spice library that this module is based on. I’ve compiled the FORTRAN version of it so that I can call it from python, and my current plan is to compare this module to the original FORTRAN function until I’ve satisfied myself that it’s working properly.

Assuming I get this module to match `prop2b` from the Spice library exactly, the question then becomes how can it be tested using pure python? Would pasting example results generated by `prop2b` into the tests be sufficient? Do you know of any other sources of 2-body propagation sample data?

brandon-rhodes:
I'm happy to hear you might have the chance to make progress!

I recommend pasting in the numeric vectors into your test, and comparing them to the new Skyfield output you produce. Here's an example, the tests I wrote over the past 2 weeks as I implemented Moon surface locations; note how I include commented-out inline SpiceyPy code that can print the matrices again whenever I want to double-check them:

https://github.com/skyfielders/python-skyfield/blob/746d8379779511dbdbe2daba5ebba7869ed11ca0/skyfield/tests/test_planetarylib.py#L9

I did have the problem that NumPy was not by default printing enough decimal places for me to know to high precision whether my results agreed with theirs. I found, as you can see, that Python’s own `repr()` was an utterly reliable way to get floats printed out at perfect resolution, so that all the precision of the original result was available for the test.

brandon-rhodes:
@JoshPaterson —

I have good news! Your propagation code here agrees with HORIZONS to a sub-millimeter level when asked the position of 1 Ceres. You mentioned above that the problem with getting high agreement is that HORIZONS adjusts the orbit for the gravitational effects of other bodies. But I checked, and their interface is able to return raw orbital parameters, not just positions, following those adjustments.

So I asked for the orbital elements of 1 Ceres on a particular date, and then for the x,y,z coordinates, and wrote a Python script that uses the first to generate the latter using your new `KeplerOrbit` class. In the comment at the top you can see the orbital parameters and position as printed raw by HORIZONS:

```
from skyfield.units import Distance, Angle

"""
2458886.500000000 = A.D. 2020-Feb-07 00:00:00.0000 TDB 
 EC= 7.705857791518426E-02 QR= 2.555508368946362E+00 IN= 2.718528770987308E+01
 OM= 2.336112629072238E+01 W = 1.328964361683606E+02 Tp=  2458240.226649156772
 N = 2.139189800548039E-01 MA= 1.382501360489816E+02 TA= 1.437265967168744E+02
 A = 2.768873850275102E+00 AD= 2.982239331603843E+00 PR= 1.682880125493173E+03

2458886.500000000 = A.D. 2020-Feb-07 00:00:00.0000 TDB 
 X = 1.334875927366032E+00 Y =-2.239607658161781E+00 Z =-1.328895183461897E+00
 VX= 8.679572171334722E-03 VY= 4.382113869938210E-03 VZ= 2.971677617262620E-04
 LT= 1.690136005998568E-02 RG= 2.926379779281297E+00 RR= 4.705579815071954E-04
"""

from skyfield.api import load
from skyfield.keplerlib import KeplerOrbit

ts = load.timescale(builtin=True)
t = ts.tdb_jd(2458886.500000000)

a = 2.768873850275102E+00 # A
e = 7.705857791518426E-02 # EC
p_au = a * (1 - e ** 2)  # Wikipedia

k = KeplerOrbit.from_mean_anomaly(
    p=Distance(au=p_au),  # see above
    e=e,  # EC
    i=Angle(degrees=2.718528770987308E+01),  # IN
    Om=Angle(degrees=2.336112629072238E+01),  # OM
    w=Angle(degrees=1.328964361683606E+02),  # W
    M=Angle(degrees=1.382501360489816E+02),  # MA
    epoch=t,  #?
    mu_km_s=None,
    mu_au_d=2.9591220828559093E-04,
    center=None,
    target=None,
    center_name=None,
    target_name=None,
)
print(k)
r, v, *rest = k._at(t)
print(r)

#eph = load('de421.bsp')
eph = load('de430t.bsp')
sun = eph['sun']
print(t.utc_jpl())
print(sun.at(t).position.au)
print(r + sun.at(t).position.au)
diff = r + sun.at(t).position.au - [
    1.334875927366032E+00,
    -2.239607658161781E+00,
    -1.328895183461897E+00,
]
print(diff)
print('difference in meters:', Distance(au=diff).m)
```

The happy result:

```
...
difference in meters: [-2.98956601e-04 -2.65739201e-04  3.32174001e-05]
```

I will accordingly plan to sit down and do a code review of this whole PR today or else this weekend. After a bit of API tweaking, I think it will be ready to land!

brandon-rhodes:
Given how long this PR sat idle before we discovered that its results are a perfect match for HORIZONS, I have decided just to merge it and then to make the tweaks in `master` that normally I would have us iterate on here. Thanks for the contribution!

PR #259, opened 2019-05-29
Check if mu is zero in OsculatingElements.__init__()

JoshPaterson:
Raises a ValueError instead of having a mysterious numpy divide by zero warning later on

### DIFF
--- a/skyfield/elementslib.py
+++ b/skyfield/elementslib.py
@@ -53,6 +53,8 @@ class OsculatingElements(object):
 
     """
     def __init__(self, position, velocity, time, mu_km_s):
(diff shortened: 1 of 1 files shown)


skyfield/elementslib.py
```diff
@@ -53,6 +53,8 @@ class OsculatingElements(object):
 
     """
     def __init__(self, position, velocity, time, mu_km_s):
+        if mu_km_s == 0: raise ValueError('`mu_km_s` is zero')
+
```
brandon-rhodes: Could you put the `raise` on its own line ­— and maybe explain in the message why zero isn't a good value, since some callers might not know why? Thanks!

JoshPaterson: Sure!

PR #264, opened 2019-06-18
Make EarthSatellite target attribute the NAIF id

JoshPaterson:
According to the [SPICE toolkit documentation](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/req/naif_ids.html#Earth%20Orbiting%20Spacecraft.):

>If an Earth orbiting spacecraft lacks a DSN identification code, the NAIF ID is derived from the tracking ID assigned to it by NORAD via:
>
>      NAIF ID = -100000 - NORAD ID code
>
>For example, NORAD assigned the code 15427 to the NOAA 9 spacecraft. This code corresponds to the NAIF ID -115427.

I confirmed that the satellite numbers in TLEs are the NORAD ID code by checking in [this file from celestrak](https://celestrak.com/NORAD/elements/noaa.txt) that the satellite number for NOAA 9 is in fact 15427.

I wasn't sure if you would want a test or not, so I included one just in case.

### DIFF
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -92,7 +92,7 @@ class EarthSatellite(VectorFunction):
         self.model = sat
         self.epoch = ts.utc(sat.epochyr, 1, sat.epochdays)
 
-        self.target = -100_000 - self.model.satnum
+        self.target = -100000 - self.model.satnum
         self.target_name = 'Satellite{0} {1}'.format(
             self.model.satnum,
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
I have held off on this for too long, probably because there's no way to know if this will break idiosyncratic user code that is using strings or objects for `self.model.satnum`. But on balance, I like the idea of a move to more standard behavior, so let’s merge and see if anyone reports problems!

PR #275, opened 2019-07-27
Allow lines not starting with 1  or 2 as TLE name.

interplanetarychris:
More permissive check than 24 character Celestrak names, allowing finding EarthSatellite Class to include descriptive names from other TLE sources.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -458,11 +458,11 @@ def parse_tle(fileobj):
             b2.startswith(b'2 ') and len(b2) >= 69):
 
             b0 = b0.rstrip(b'\n\r')
-            if len(b0) == 24:   # Celestrak
-                name = b0.decode('ascii').rstrip()
-                names = [name]
-            elif b0.startswith(b'0 '): # Spacetrack 3-line format
+            if b0.startswith(b'0 '): # Spacetrack 3-line format
                 name = b0[2:].decode('ascii').rstrip()
+                names = [name]
+            elif ((not b0.startswith(b'1 ')) and (not b0.startswith(b'2 '))): # Celestrak and others
+                name = b0.decode('ascii').rstrip()
                 names = [name]
             else:
                 name = None
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -48,6 +48,8 @@ class EarthSatellite(VectorFunction):
     ``epoch``
         A Skyfield :class:`~skyfield.timelib.Time` giving the exact
         epoch moment for these satellite orbit parameters.
+    ``name``
+        Satellite name
 
     If you are interested in the catalog entry details, the SGP4 model
     parameters for a particular satellite can be accessed through its
@@ -55,6 +57,10 @@ class EarthSatellite(VectorFunction):
 
     ``model.satnum``
         The unique satellite NORAD catalog number given in the TLE file.
+    ``model.classification``
+        Satellite classification or `U`
+    ``model.intldesg``
+        International designator
     ``model.epochyr``
         Full four-digit year of this element set's epoch moment.
     ``model.epochdays``
@@ -67,6 +73,10 @@ class EarthSatellite(VectorFunction):
         Second time derivative of the mean motion (ignored by SGP4).
     ``model.bstar``
         Ballistic drag coefficient B* in inverse earth radii.
+    ``model.ephtype``
+        Ephemeris type (ignored by SGP4 as determination now automatic)
+    ``model.elnum``
+        Element number
     ``model.inclo``
         Inclination in radians.
     ``model.nodeo``
@@ -77,8 +87,10 @@ class EarthSatellite(VectorFunction):
         Argument of perigee in radians.
     ``model.mo``
         Mean anomaly in radians.
-    ``model.no``
+    ``model.no_kozai``
         Mean motion in radians per minute.
+    ``model.revnum``
+        Revolution number at epoch [Revs]
 
     """
     center = 399


brandon-rhodes:
Thanks for your interest in improving the satellite interface!

1. Is there a source of TLE files I can look over whose satellite names this new approach would detect? It looks like any random text on the line before the TLE would now be assumed to be the satellite name, which the old code tries to avoid in case TLE data is embedded in a larger text. But maybe the penalty of sometimes pulling in a random line of text as the satellite name is not as large as I'm imagining? I had tried to imagine a beginner's reaction if they printed a dictionary of satellites and saw, say, the final line of a paragraph as a one of the dictionary keys. (I probably should never have offered satellites as a dictionary in the first place, and instead have shown sample code for folks to build dictionaries who needed them; then a mis-assigned name wouldn't be as bad?)

2. It looks like an unrelated commit has been added — did you intend to instead add it to your other pull request, since it looks like documentation for new satellite fields?

interplanetarychris:
-Chris
@interplanetary

> Thanks for your interest in improving the satellite interface!
> 
There will be more coming!  I’m working on a project which will make heavy use of both python-(skyfield, SGP4).  

> Is there a source of TLE files I can look over whose satellite names this new approach would detect? It looks like any random text on the line before the TLE would now be assumed to be the satellite name, which the old code tries to avoid in case TLE data is embedded in a larger text. But maybe the penalty of sometimes pulling in a random line of text as the satellite name is not as large as I'm imagining? I had tried to imagine a beginner's reaction if they printed a dictionary of satellites and saw, say, the final line of a paragraph as a one of the dictionary keys. (I probably should never have offered satellites as a dictionary in the first place, and instead have shown sample code for folks to build dictionaries who needed them; then a mis-assigned name wouldn't be as bad?)
> 
> 
The easiest one would be:
https://www.prismnet.com/~mmccants/tles/index.html <https://www.prismnet.com/~mmccants/tles/index.html>

Specifically: https://www.prismnet.com/~mmccants/tles/classfd.zip <https://www.prismnet.com/~mmccants/tles/classfd.zip>

There are also the variants of “line 0” which sometimes include the leading “0 “ and sometimes just have a name.
> 
> It looks like an unrelated commit has been added — did you intend to instead add it to your other pull request, since it looks like documentation for new satellite fields?
> 
> 
Yes - I’m new to GIT.  I didn’t think enough to separate out the commits in different branches, and the same code ended up in the Vallado update I put in (no changes between the two).  Let me know if it would be much better to separate them, and I can re-submit.

ckuethe:
You can also get non-leading-zero TLEs from https://tle.info/

As for line length limits, I'd say cap it at 69 characters, just like lines `1` and `2` of a TLE. 
Space-Track says optional line 0 is [24 characters](https://www.space-track.org/documentation#/tle) and if you can make their API include or skip the leading 0 if you want. I haven't looked to see if they have satellites whose reported name is truncated if you ask for the leading 0. [This NASA page](https://spaceflight.nasa.gov/realdata/sightings/SSapplications/Post/JavaSSOP/SSOP_Help/tle_def.html) says line 0 is only 11 characters.

Since this line is not necessary for computing orbits, and there is already some discrepancy in the definition of the line length I figure it's fair game to make it as long as the other lines to let users put some more metadata in there, eg. observer name, site, object name, orbit shape. This is only 68 characters: `ckuethe,99999,2019-09-24T22:30:17,unknown,2019-999ZZ,2048x32768,51.2`

brandon-rhodes:
So it took me a couple of months to realize why I am wary of this approach.

I thought at first it was its fragility with respect to general purpose text files. If an email says `Here are the elements:` then has the two lines of satellite data, then doing a `parse_tle()` with this PR merged would think that `Here are the elements:` was the name of the satellite. The strict check for 24 characters is supposed to make this unlikely (except where the previous line of the email text just happens to be 24 characters!).

But I think the real reason is that I am not sure I like how I designed `parse_tle()`. It's a complicated routine that has switches to try to automatically detect and handle different cases. Why didn’t I just write two much simpler routines, one called `parse_celestrak_tle()` and `parse_spacetrack_tle()`, that each specialized more cleanly in one “brand” of TLE file?

I think that, even though I wasn’t clever enough at the time to do that myself, we should head in that direction. Could you rewrite this PR as its own new kind of parsing routine that knows how to interpret exactly the kind of TLE you’re working on? Then we can, after the PR is merged, start migrating the Skyfield docs towards encouraging folks to select and use the right parsing routine, instead of having one big magical parsing routine that tries and fails to do the right thing for satellite elements in plain text files and also for satellite elements in dedicated TLE files with nothing else in them.

Does that sound doable?

interplanetarychris:
I don’t know that there’s a prevailing “standard” for line 0 / name field, other than it not be longer than the other data lines.

I’ve written what is intended to be a rigorous and comprehensive TLE-parsing and validation routine in https://github.com/consensys-space/trusat-orbit/blob/master/tle_util.py <https://github.com/consensys-space/trusat-orbit/blob/master/tle_util.py>

Also included are functions for going from SGP4 variable-space to TLE variable space such as 
make_tle_from_SGP4_satrec()

We can discuss if it makes sense to use those functions in the PR, or do something separate/independent of the additional dependency.

-Chris
@interplanetary

brandon-rhodes:
Thanks for the link to your own routine! I am finishing up a talk this week for code::dive next week, but I deliver it Wednesday morning and will be able to return my attention to Skyfield. If I don't follow up here with further thoughts within the next few days, expect something in the second half of next week!

brandon-rhodes:
Happily, I did wind up taking this advice when I went in and rewrote that routine recently: it now accepts whatever text is above the TLE lines as the name. I decided that my objection above — that users would be confused to see non-name lines of text pulled in — was pretty unlikely. Thanks for the idea!

PR #276, opened 2019-07-29
Sgp4lib update/7 dec 15 vallado version

interplanetarychris:
Bring current with python-sgp4 [pull request 35.](https://github.com/brandon-rhodes/python-sgp4/pull/35)

Documentation update to reflect change from model.no to model.no_kozai, 
as well as expose all other model variables available from the TLE.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -458,11 +458,11 @@ def parse_tle(fileobj):
             b2.startswith(b'2 ') and len(b2) >= 69):
 
             b0 = b0.rstrip(b'\n\r')
-            if len(b0) == 24:   # Celestrak
-                name = b0.decode('ascii').rstrip()
-                names = [name]
-            elif b0.startswith(b'0 '): # Spacetrack 3-line format
+            if b0.startswith(b'0 '): # Spacetrack 3-line format
                 name = b0[2:].decode('ascii').rstrip()
+                names = [name]
+            elif ((not b0.startswith(b'1 ')) and (not b0.startswith(b'2 '))): # Celestrak and others
+                name = b0.decode('ascii').rstrip()
                 names = [name]
             else:
                 name = None
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -48,6 +48,8 @@ class EarthSatellite(VectorFunction):
     ``epoch``
         A Skyfield :class:`~skyfield.timelib.Time` giving the exact
         epoch moment for these satellite orbit parameters.
+    ``name``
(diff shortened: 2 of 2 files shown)


skyfield/iokit.py
```diff
@@ -458,12 +458,12 @@ def parse_tle(fileobj):
...
 
             b0 = b0.rstrip(b'\n\r')
-            if len(b0) == 24:   # Celestrak
-                name = b0.decode('ascii').rstrip()
-                names = [name]
-            elif b0.startswith(b'0 '): # Spacetrack 3-line format
+            if b0.startswith(b'0 '): # Spacetrack 3-line format
                 name = b0[2:].decode('ascii').rstrip()
                 names = [name]
+            elif ((not b0.startswith(b'1 ')) and (not b0.startswith(b'2 '))): # Celestrak and others
```
brandon-rhodes: Could you combine these into a single `startswith()` to which you provide both strings?

skyfield/iokit.py
```diff
@@ -458,12 +458,12 @@ def parse_tle(fileobj):
             b2.startswith(b'2 ') and len(b2) >= 69):
 
             b0 = b0.rstrip(b'\n\r')
-            if len(b0) == 24:   # Celestrak
-                name = b0.decode('ascii').rstrip()
-                names = [name]
-            elif b0.startswith(b'0 '): # Spacetrack 3-line format
+            if b0.startswith(b'0 '): # Spacetrack 3-line format
                 name = b0[2:].decode('ascii').rstrip()
                 names = [name]
```
brandon-rhodes: Just for completeness, probably check whether the `name` has a nonzero number of characters first?

skyfield/iokit.py
```diff
@@ -458,12 +458,12 @@ def parse_tle(fileobj):
...
+            elif ((not b0.startswith(b'1 ')) and (not b0.startswith(b'2 '))): # Celestrak and others
+                name = b0.decode('ascii').rstrip()
+                names = [name]
```
brandon-rhodes: The `name` could be the empty string here, if a blank line (or whitespace) preceded the two TLE lines, so please add a check for whether `name` is nonempty. Thanks!

brandon-rhodes:
Happily, the changes to TLE parsing wound up happening in a big simplification when I rewrote and replaced that routine, so thanks for the idea, I ultimately decided it was a good one. That left only the attribute description updates, which I've just merged. Thanks! I'm not sure why it took me a few weeks to get back to this PR and notice it was nearly complete; I guess I spend too much time over in the Issues. :)

Issue #115, opened 2016-10-20
Limited satellite information compared to pyephem


dmopalmer:
In pyephem, you can ask when a satellite will rise, set, and transit.  Also whether it never sets or rises.

Are those functions unavailable in the current version of skyfield?  Not-yet-implemented or never-will-be?

eheinrich:
Same question for finding out if a satellite is eclipsed.

brandon-rhodes:
@dmopalmer Great question! They are merely not-yet-implemented. I definitely want to see Skyfield grow to do everything PyEphem did and more. For things like rising and setting, I am not yet sure whether it will be best to implement our own API or instead have instructions in the documentation showing people how to use the existing solvers in SciPy and elsewhere in the ecosystem to find conditions like rising and setting. One hope I have had as I have been writing Skyfield is that its tight integration with NumPy will make lots of tasks that had to be implemented by hand in PyEphem will in the modern science ecosystem become trivial — "grab the right solver off the shelf and let it run!"

Let's keep this issue open until we have satellite rising and setting solutions that we all agree are at least as good as PyEphem's!

Here's the Jupyter Notebook that I created back in 2014 when I last had the time to sit down and start investigating what satellite risings and settings look like:

https://github.com/skyfielders/astronomy-notebooks/blob/master/Solvers/Earth-Satellite-Passes.ipynb

If that work can be expanded, then we could put together a good story for finding passes!

dmopalmer:
I am working on this now, making it an almanac function:
```
def find_satellite_events(start_time: Time, end_time: Time,
                          satellite: EarthSatellite, topos: Topos,
                          horizon: float = 0):
...

SATELLITE_EVENTS = [
    'rise',  # rises above user-defined horizon
    'culminate',  # reaches highest point during pass.  (Not the same as meridian crossing)
    'set',  # sets below user-defined horizon
]
```
Questions

1. What python versions are we maintaining compatability for?
2. scipy is not in the current requirements.  I need an optimizer to search for maxima and roots.  Is it OK to add scipy to the requirements?
3. What is the standard for argument type annotations?  Just don't have them?
4. Helper function:
```
def linspace_time(start_time, stop_time, *, num=50, step=None, endpoint=True)
```
does this already exist for skyfield.api.Times ?  It seems generally useful.  Should it go in some utility file instead of being hidden under almanac?

brandon-rhodes:
[...]

PR #281, opened 2019-08-18
Earth satellites

Imperator26:
This PR implements a method that allows to get the rise time, the set time, the apex time and the altitude at the apex for any man-made satellite orbiting the Earth. I took most of the code from an article on https://rhodesmill.org I believe, but I can't seem to find it. I adapted the code months ago but decided to propose it only now. Please link to it if you're able to.

This function requires the optimize() method from the SciPy lib.

I made this PR so that we can further discuss this implementation. Perhaps it could be generalized to any planet and therefore to any satellite, man-made or not.


brandon-rhodes:
Interesting! I'll have to find time soon to try this routine out.

Skyfield's traditional claim is that it has no binary dependencies besides NumPy, and given that SciPy is a very large dependency — about twice the size of NumPy — I'm inclined to avoid adding it has a hard dependency of Skyfield; my impression is that some folks on small machines (think of small boards running Python, for example) already have trouble just getting NumPy and an ephemeris downloaded.

But we should come up with a good story where things would fit that require more dependencies than Skyfield itself. I do have some code that requires Pandas in Skyfield, that when run asks for Pandas to be installed.

For something like finding a maximum, I'm surprised that it requires SciPy; I already have the scaffolding of a maxima-finder in Skyfield, I wonder if we could get that routine working instead?

Imperator26:
Yeah I was aware of the issues regarding adding SciPy to the dependencies, as a metter of fact this was the main reason why I decided to not make a PR when I adapted the code in the first place.

I might have some time to take a look at the maxima-finder already in the lib. I'll post any update here and fix the code if I'm successful.

Imperator26:
I have started editing the code to support the `find_discrete()` method over SciPy's optimize. At this point `get_satellite_passes()` returns only the rise time and the set time. I'll improve it in the near future. Compared to the method with SciPy this one is much faster.

Imperator26:
The code is pretty much done. Although I don't think that `almanac.py` is the right file where `get_satellite_passes()` should be put in.

skyfield/almanac.py
```diff
@@ -135,6 +144,134 @@ def _find_maxima(start_time, end_time, f, epsilon=EPSILON, num=12):
...
+            ephemeris,
+            topos,
+            satellite,
+            t
+        )
+        for t in apices_time
+    ]
+
+    return list(
+        zip(
```
brandon-rhodes: Wow — does this work even if, say, the start time falls inside the time that the satellite is visible? I guess I should try it out with some sample values after the PR lands.

skyfield/timelib.py
```diff
@@ -461,6 +479,17 @@ def utc_datetime_and_leap_second(self):
             dt = datetime(year, month, day, hour, minute, second, milli, utc)
         return dt, leap_second
 
+    def timestamp(self):
```
brandon-rhodes: As Python 2 doesn’t support this concept, I think I prefer that we not add it to Skyfield. But I should back up and ask: what are the advantages of datetime timestamp values? What use case would be helped by bringing them into Skyfield? I've not yet seen any astronomy code that uses them.

skyfield/almanac.py
```diff
@@ -230,9 +370,85 @@ def moon_phase_at(t):
...
+
 def _distance_to(center, target):
     def distance_at(t):
         t._nutation_angles = iau2000b(t.tt)
         distance = center.at(t).observe(target).distance().au
         return distance
     return distance_at
+
+
+def satellite_visible(ephemeris, topos, satellite, alt_deg_thresh=0):
```
brandon-rhodes: These are a useful collection of routines! I wonder if they go beyond what an "almanac" is, though. Might they belong in a module of their own? For example, we could put them in the sgp4 module as one possible place for them to live?

brandon-rhodes: (And we don't need to answer that immediately — I can always shuffle things around before the next release, after accepting this PR — but I thought I'd ask to hear your ideas on the question!)

Imperator26:
No worries. I'll try to address your comments as soon as possible

skyfield/almanac.py
```diff
@@ -39,6 +46,7 @@ def fraction_illuminated(ephemeris, body, t):
     a = phase_angle(ephemeris, body, t).radians
     return 0.5 * (1.0 + cos(a))
 
+
```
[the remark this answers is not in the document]
Imperator26: I'm fine with a single blank line, but in general I try to follow PEP8 conventions

skyfield/almanac.py
Imperator26: Yeah I thought about that too. I didn't really know where to put the code so I decided for the closest one in terms of tasks

skyfield/timelib.py
Imperator26: I just noticed that find_discrete() should work fine with Time objects. If that's the case then the entire code could be slimmed down quite a bit

skyfield/almanac.py
Imperator26: I think it should

skyfield/almanac.py
```diff
@@ -135,6 +144,134 @@ def _find_maxima(start_time, end_time, f, epsilon=EPSILON, num=12):
...
+    Returns a list of tuples each of which is of dimension 1x7 and contains,
+    respectively, the rising Time, the apex Time, the setting Time, the rising
+    azimut, the azimut at the apex, the setting azimut in degrees and the
+    altitude at the apex. The rising Time and the setting Time are based on the
+    set alt_deg_threshold.
+    """
+
+    # Get timestamps
+    if isinstance(from_t, Time):
```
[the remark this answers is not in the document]
Imperator26: Makes sense

skyfield/almanac.py
Imperator26: It should be checked out anyway though. If the arrays don't match the extra elements get dropped and rise times get paired with previous rise_times

For example:
if rise_time = [rt2, rt3, ..., rtn]
and set_time = [st1, st2, ..., stn]

rt1 has passed so when zipping rt2 is paired with st1, rt3 with st2 and so on.

Perhaps the arrays should still be zipped in the end but the return could look like
```
return list(
    zip(
      rise_time,
      azimuts_at_rise_time
    )
  ),
  list(
    zip(
      apices_time,
      azimuts_at_apex_time,
      altitudes
    )
  ),
  list(
    zip(
      set_time,
      azimuts_at_set_time
    )
  )
```

skyfield/almanac.py
```diff
@@ -135,6 +144,134 @@ def _find_maxima(start_time, end_time, f, epsilon=EPSILON, num=12):
...
+        satellite_azimut_at(
+            ephemeris,
+            topos,
+            satellite,
+            t
+        )
+        for t in set_time
+    ]
+
+    rise_set_times = list(zip(rise_time, set_time))
```
Imperator26: > It should be checked out. If the arrays don't match the extra elements get dropped and rise times get paired with previous rise_times
> 
> For example:
> if rise_time = [rt2, rt3, ..., rtn]
> and set_time = [st1, st2, ..., stn]
> 
> If the satellite has already risen then it means that rt1 has passed so when zipping, rt2 is paired with st1, rt3 with st2 and so on.
> 
> Perhaps the arrays should still be zipped in the end but the return could look like
> 
> ```python
> return list(
>     zip(
>       rise_time,
>       azimuts_at_rise_time
>     )
>   ),
>   list(
>     zip(
>       apices_time,
>       azimuts_at_apex_time,
>       altitudes
>     )
>   ),
>   list(
>     zip(
>       set_time,
>       azimuts_at_set_time
>     )
>   )
> ```

Same thing here

skyfield/timelib.py
Imperator26: A lot of time has passed. I thought I added the timestamp methods to satisfy the parameters of find_discrete(), but I was wrong. I believe I added them for simplicity or because the project I was working on at the time was using timestamps and datetime but I don't remember exactly

brandon-rhodes: In that case, let’s definitely go for slimming the code down and making it simpler, and later if folks want to call this routine with timestamps, we can add documentation showing them how to do the conversion themselves. Thanks!

Imperator26:
> Thanks for the work you put into this PR! I am interested in adding this code to Skyfield. Could you add a basic test that shows it working for a given satellite? Also, I have made a few comments in the code; feel free to answer when you see that I have misunderstood something or am asking the wrong question.
> 
> My answer was delayed because of recent travel, but I will hopefully be more responsive through the rest of the year. It must have been frustrating to wait this long for a review.

Where should I be putting the example?

satellite_passes.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Load timescale
+ts = load.timescale()
+
+# Load planets, get Earth and set the ground station
+ephemeris = load("de421.bsp")
+ground_station = Topos('44 N', '11 E', elevation_m=70)
+
+# Load satellites
+satellites = load.tle("http://celestrak.com/NORAD/elements/weather.txt")
```
ckuethe: What if I don't have internet, or what if NOAA18 suddenly gets hit by an asteroid? I love to write code when I'm on an airplane where there are fewer interruptions? There is already a hard coded TLE for the ISS in https://github.com/skyfielders/python-skyfield/blob/master/skyfield/tests/test_earth_satellites.py#L9 - maybe reuse that?

ckuethe:
Sorry about my derpy comment on your example usage, I meant to put that here.

Anyway, I think https://github.com/skyfielders/python-skyfield/blob/master/skyfield/tests/test_earth_satellites.py would be a good place. There is already a TLE for the ISS there, along with some similar time-based calculations (checking the ISS against where Horizons said it was at that point in time).

Imperator26:
> Sorry about my derpy comment on your example usage, I meant to put that here.
> 
> Anyway, I think https://github.com/skyfielders/python-skyfield/blob/master/skyfield/tests/test_earth_satellites.py would be a good place. There is already a TLE for the ISS there, along with some similar time-based calculations (checking the ISS against where Horizons said it was at that point in time).

`satellite_passes.py` is more of an example than a test. I'm not checking the results of the script in any way.

dmopalmer:
I didn't see this and independently implemented it as pull request #321 on #115 .  (My version also uses scipy).

brandon-rhodes:
> Imperator26 requested a review from brandon-rhodes on Nov 13, 2019

Ack! I'm two months late! How did I not see this on my GitHub notifications? I hope I didn't open the link (merely clicking on it dismisses the notification), then accidentally close the tab. Maybe the laptop ran out of power and the tab didn't come back when I rebooted; that happens sometimes.

I'll read both pull requests again right now, and see where things stand.

brandon-rhodes:
So, due to my losing track of this Pull Request over the holidays,
it unfortunately lingered for 8 weeks without further feedback —
long enough for a rival pull request to appear.

(It’s interesting that after so many years without this feature,
two fully worked out solutions appeared within a few months of each other!)

Having read over both of them,
I’m in favor of keeping them both:
they both have good ideas that need to be part of Skyfield!

Here are several good features this PR, #281:

* Uses Skyfield’s vector-aware search routines
  that uses a single loop to find all _n_ events over a time period,
  instead of having to commence a separate search for each one.
* Uses vector math to handle results like “all rise times” and
  “all set times” as single objects.
* Following the example of all other Skyfield code,
  carefully names `alt_deg_thresh` so that the name explicitly states
  what unit it should be provided in.

While here are several good features of #321:

* Generates positions very efficiently
  by computing the instantaneous difference
  between the position of the Topos and the position of the satellite.
* Documentation is included.
* Two tests are included! And they are thoughtfully constructed
  to challenge the algorithm under hand-picked circumstances.
  I do worry a little about their runtime,
  but have not tried them yet to see;
  if they slow down testing then I might ask
  whether we could get roughly the same guarantees with fewer operations.

So here are the steps I plan to take.

1. Merge #321,
   because that will instantly install documentation and tests
   for satellite finding into Skyfield.
2. By hand,
   take the ideas here in #281 and use them
   to keep the satellite tests working
   while gradually removing the dependency on SciPy.

I will keep this pull request open until I am done with both steps.

Thanks again very much to both of you for providing the raw materials
which will finally let this long-desired feature appear in Skyfield!

Bernmeister:
Whilst you have the thing up on the blocks, satellite passes are great, but visible passes are even better!  

I've used the code inspired by https://space.stackexchange.com/questions/4339/calculating-which-satellite-passes-are-visible to do this in PyEphem...not sure if it helps eventually with Skyfield.

dmopalmer:
I agree.  I had planned to add this eventually to my pull, but decided to do only the first part initially.  The values are commented out in the SATELLITE_EVENTS list (almanac.py:341)

The steps to determine when a satellite is illuminated are:
1. Write a function that determine the ratio of a) The Earth-Sat-Sun angle to b) The Earth's apparent angular radius as seen from the satellite.  When this function is less than unity, the satellite is eclipsed; greater than unity means illuminated.
2. Find the times where a)/b) crosses unity.  This requires finding the sequence of minima and maxima, and then finding the transitions where a minimum-maximum (or vice-versa) sequence crosses unity, and search in those intervals for the crossing point.

You don't want to do 2. just by searching for discrete transitions without the extrema step because you are likely to miss them.

Once you have that search routine you can apply it to the time intervals when the satellite is above the local horizon during observer's local night (by what twilight definition?).

If the a)/b) ratio is near unity, then the satellite may be in penumbra, or it may be illuminated by refracted light (sunset glow) and still visible, but that is a more detailed specification.

Imperator26:
> Whilst you have the thing up on the blocks, satellite passes are great, but visible passes are even better!
> 
> I've used the code inspired by https://space.stackexchange.com/questions/4339/calculating-which-satellite-passes-are-visible to do this in PyEphem...not sure if it helps eventually with Skyfield.

Well, they kinda serve to different purposes so they should be both implemented

brandon-rhodes:
Thanks for the note, yes, I agree that visibility is an important additional calculation! If you have a moment, open an Issue to track it.

I suspect we'll want visibility to be a separate computation, that takes the output of the plain "passes" routine as input, and filters those passes down to the periods during which the satellite is illuminated.

dmopalmer:
More thoughts on _find_maxima and how it can be used in this:

1. For #321, the code can be hugely simplified by using your `_find_maxima`.  However it needs a `max_first_step` parameter to initially search on a sub-orbit timescale and avoid jumping over the maxima.
2. Using `np.positive` instead of `np.sign` solves the problem of equal values at gridpoints.  (Change the 2 to 1 in the test)
3. Looking for a minimum becomes a matter of flipping the sign in that search.  Looking for extrema becomes a matter of looking for non-zero.  So you can easily extend the function to find minima, maxima, or both.

If I had realized that the function returned multiple maxima, I would have made these changes and used it myself in the #321 pull.

You could gain a bit of efficiency by expending more effort on the gridpoint spacing on each step: having closer spacing where you expect the maximum to be based on e.g. a parabolic interpolation.  But I doubt that is worth it.

Imperator26:
> So, due to my losing track of this Pull Request over the holidays,
> it unfortunately lingered for 8 weeks without further feedback —
> long enough for a rival pull request to appear.
> 
> (It’s interesting that after so many years without this feature,
> two fully worked out solutions appeared within a few months of each other!)
> 
> Having read over both of them,
> I’m in favor of keeping them both:
> they both have good ideas that need to be part of Skyfield!
> 
> Here are several good features this PR, #281:
> 
> * Uses Skyfield’s vector-aware search routines
>   that uses a single loop to find all _n_ events over a time period,
>   instead of having to commence a separate search for each one.
> * Uses vector math to handle results like “all rise times” and
>   “all set times” as single objects.
> * Following the example of all other Skyfield code,
>   carefully names `alt_deg_thresh` so that the name explicitly states
>   what unit it should be provided in.
> 
> While here are several good features of #321:
> 
> * Generates positions very efficiently
>   by computing the instantaneous difference
>   between the position of the Topos and the position of the satellite.
> * Documentation is included.
> * Two tests are included! And they are thoughtfully constructed
>   to challenge the algorithm under hand-picked circumstances.
>   I do worry a little about their runtime,
>   but have not tried them yet to see;
>   if they slow down testing then I might ask
>   whether we could get roughly the same guarantees with fewer operations.
> 
> So here are the steps I plan to take.
> 
> 1. Merge #321,
>    because that will instantly install documentation and tests
>    for satellite finding into Skyfield.
> 2. By hand,
>    take the ideas here in #281 and use them
>    to keep the satellite tests working
>    while gradually removing the dependency on SciPy.
> 
> I will keep this pull request open until I am done with both steps.
> 
> Thanks again very much to both of you for providing the raw materials
> which will finally let this long-desired feature appear in Skyfield!

I'll start working on this more seriously next week

brandon-rhodes:
> I'll start working on this more seriously next week

Happily, there's no need! I'm most of the way done with getting the core of your approach installed into the satellite routines from the other Pull Request. If you'll hang on another day or two, I plan to finish with my work, and then you can take a look on master and see if you like the result or want to suggest further changes.

Imperator26:
> > I'll start working on this more seriously next week
> 
> Happily, there's no need! I'm most of the way done with getting the core of your approach installed into the satellite routines from the other Pull Request. If you'll hang on another day or two, I plan to finish with my work, and then you can take a look on master and see if you like the result or want to suggest further changes.

Alright

brandon-rhodes:
Okay! There might be a couple of more edge cases hiding, and there are a few more approaches I want to experiment with, but I have a routine working that I will be happy to go ahead and release first — could anyone try it out and let me know how it behaves? I think these two steps will work:

    pip install https://github.com/brandon-rhodes/python-jplephem/archive/master.zip
    pip install https://github.com/skyfielders/python-skyfield/archive/master.zip

You should find that satellites now have a `.find_passes(topos, t0, t1)` method that returns an array of times and an array of 0=rise 1=culminate 2=set event labels.

I was happy to see that it caught a culmination that previous algorithms had not. I don't think there's a guarantee it will catch all such events, but here's satellite Ariane 5B (44802) on January 5th viewed from '42.3581 N', '71.0636 W':

image

I thought for a moment it was a bug when I saw Skyfield say:

```
 A.D. 2020-Jan-05 10:40:12.4511 UT 0
 A.D. 2020-Jan-05 13:03:31.1903 UT 1
 A.D. 2020-Jan-05 19:28:31.8061 UT 1
 A.D. 2020-Jan-05 19:44:53.5133 UT 2
```

But, happily, it's just a behavior of the satellite.

Anyway, if anyone could try it out, I could consider a release of the feature. If I can convince myself not to stop and do a huge rewrite so that it would keep, rather than throwing out, all the other data it knows about where the satellite was at each point. But if I dive immediately into that, I'm not sure how long it might take!

dmopalmer:
It is probably not worth it to store all the satellite information at each timestep you calculate during the search for pass events..  

The user might request altitude/azimuth/etc. at the times of rise/culminate/set so that those will have to be recalculated, but those three points compared to the large number of points you have to process to do the entire search.  Instead the user might do a plot of position every 10 seconds, and those 10 second points are unlikely to be identical to the points generated in the search.

dmopalmer:
It would be better (less unexpected, less arbitrary) if `minimum_altitude_degrees` defaulted to 0 rather than 20.0 degrees.

The documentation for `find_passes()` should point out that multiple-culminations between rise and set are possible.  (Simplest example is a non-circular 24-hour orbit where the satellites moves relative to a geostationary point, but never rises or sets.)

The docstring for `find_passes()` says that it returns a times,altitudes tuple instead of times and event labels.

Finding the times when it is sunlit/eclipsed should probably be a new issue rather than delaying this release.

Otherwise I have successfully installed and executed the code.

brandon-rhodes:
> I have successfully installed and executed the code.

Wonderful! I'm happy that the code was able to run successfully on a machine other than my own. Did you get a sense for its execution speed on your machine versus the earlier version?

> It would be better (less unexpected, less arbitrary) if `minimum_altitude_degrees` defaulted to 0 rather than 20.0 degrees.

Oh, I had copied the value from the tests without thinking about it! Thanks.

> The documentation for `find_passes()` …

Oh, yes, it's still the documentation is for another function that it no longer resembles, alas. I'll rewrite it.

> should point out that multiple-culminations between rise and set are possible.

Done!

> Finding the times when it is sunlit/eclipsed should probably be a new issue rather than delaying this release.

Agreed! It can be added as a filter for satellite passes. I suppose radio folks will want this plain output without the expense of computing whether the satellite is in sunlight.

brandon-rhodes:
Now that I'll pulled together what looks like a working implementation, I'm going to close this PR, and plan on a release hopefully by the end of the weekend. Thanks for the code you contributed here, @Imperator26, and I hope you can see your original ideas from this pull request — in particular the `find_discrete()` call — in the new `find_events()` method.

redraw:
Just came to say thanks here for the work done to add this feature (I'd missed it from PyEphem!), and @brandon-rhodes to welcome and assemble those PR.

PR #290, opened 2019-10-01
Document other dependencies, fixes #135

### LINKED ISSUE #135: sgp4 is an unlisted dependency
The website/readme gives the impression that the only dependency is numpy, although some astropy features can be used. However, it seems that the package sgp4 is required - `from skyfield.api import load` will fail with the following:
`File "/home/devans/.local/lib/python2.7/site-packages/skyfield/sgp4lib.py", line 4, in <module>`
`from sgp4.earth_gravity import wgs72`
`ImportError: No module named sgp4.earth_gravity`

I don't know if there are meant to be ways to use skyfield without sgp4, but certainly all of the example code requires skyfield.api.load to work.

ckuethe:


README.rst
```diff
@@ -32,8 +32,9 @@ The result:
    +09deg 03' 23.1"
    2.33251 au
 
-Skyfield’s only binary dependency is NumPy.
-Once that is available, Skyfield can usually be installed with::
+Skyfield's only binary dependency is NumPy though jplepehm and sgp4
+are also required. Once those are available, Skyfield can usually be
```
brandon-rhodes: “Once those are available” — I think this phrase is misleading? It sounds like they need to be on the person's system already, which the don't, they merely need to be "available" on PyPI (which, as their author, I make sure they are). It sounds to me like this is likely to make people do extra work?

requirements.txt
```diff
@@ -0,0 +1,3 @@
+jplephem>=2.3
```
brandon-rhodes: I would prefer to have only one file be in charge of what gets installed for developers, and at the moment that's `requirements.sh`. I'm not opposed to discussing whether it could, given how much healthier the Python ecosystem is these days, be replaced with a `requirements.txt`, but:

1. The file would need to do everything `requirements.sh` does.
2. And, the file should not duplicate information already in the `setup.py`. It should ask for this package "skyfield" to be installed, which will add automatically the dependencies in `setup.py`.

README.rst
ckuethe: Possibly my expectations were incorrect, but `python3 setup.py install` didn't pull down all the dependencies. Happy to change "available" to "installed".

requirements.txt
ckuethe: I'm not used to looking for `requirements.sh`, just `requirements.txt`. Possibly moving the dependencies to `requirements.txt` and then `requirements.sh` can read/use it, as well as `pip3 install -r requirements.txt`

ckuethe:
The snag, if you can call it that, was that I was surprised by needing jplephem. I wasn't surprised about sgp4 because I already had that installed.

README.rst
brandon-rhodes: The command `setup.py install` in Python means, as far as I know, “installed this package but _not_ its dependencies” — so I think the result you saw was correct, and what you would see for any other package that has dependencies? The command to install including dependencies is `pip install .`.

requirements.txt
brandon-rhodes: I'd be happy with a pull request that converted `requirements.sh` to `requirements.txt` — I have gradually stopped using conda for everything, now that wheels have become common enough.

README.rst
ckuethe: I'm sure that's what I get for not reading the docs super carefully every time I set up a new machine.

Roughly though, if I see a `requirements.txt` I `pip install -r` it, and then `setup.py` works. But I didn't see a `requirements.txt` so I didn't do anything special.

brandon-rhodes: Ah, yes, that in general won’t work. If you want to install a package from a checkout, `pip install .` installs the package and its dependencies. If you want to be able to edit files in the checkout and have them reflected in programs, `pip install -e .` will install the dependencies, then use symlinks to install the `.py` files so that edits to them in the repository are reflected in the next run of each Python program.

Glad we could get that straight for you!

brandon-rhodes:
An update:

Your questions convinced me to abandon `requirements.sh` and, in October, to switch the project back to a `requirements.txt`. Thanks to the fact that the new file ends with `-e .`, all Skyfield's direct dependencies should get installed when you `pip install -r requirements.txt`. Thanks for spurring me to go take a look and wheels, which led to me discover that `pip` is able to install binary versions of packages far more often than it was able to back when I switched to Conda and wrote `requirements.sh` at the beginning of 2015.

Since the situation should now be resolved, I'm closing this pull request, but please re-open if there are issues you're still encountering. Thanks!

PR #292, opened 2019-10-06
Expanded documentation for EarthSatellite.__init__

jesseli2002:
I've never used RST files before but hopefully my internal neural net has a good enough content-aware-fill that the syntax stays correct.


brandon-rhodes:
Thank you! I had not realized that the inline docs failed to mention the proper way to initialize.

PR #306, opened 2019-12-13
Add support for solar terms

mountain:
Solar terms are widely used in East Asia, the calculation is just the same as seasons in Skyfield

https://en.wikipedia.org/wiki/Solar_term


brandon-rhodes:
Thanks for writing this up! Ping me once you have the CI green.

Given how many different cultures exist whose astronomical traditions might also generate functions, I imagine the `almanac.py` file growing without bound and becoming an expensive import for folks on small systems or who only want numbers for one particular culture. Could you update this PR to move the routines into an `almanac_east_asia.py` module to begin the process of making the almanac code less monolithic?

To make things symmetric, I could then add an `almanac_europe.py` routine to compute cross-quarter days, which are the old solar-based holidays that May Day and Halloween are memories of, which are based on splitting the year into 8 rather than simply 4 as we do in modern times.

Thanks again for working to make this information available for Skyfield users!

mountain:
Sure, I will refactor them into almanac_east_asia.py and test_almanac_east_asia.py

mountain:
I am a little bit confused of the failure in doctest part in python3.6.

[The failure report](https://travis-ci.org/skyfielders/python-skyfield/jobs/624745043#L456) shows the previous test case on twilight is influenced by the latest test case.

brandon-rhodes:
> I am a little bit confused of the failure in doctest part in python3.6.
> [The failure report](https://travis-ci.org/skyfielders/python-skyfield/jobs/624745043#L456) shows the previous test case on twilight is influenced by the latest test case.
> Any hint to fix this?

No! Maybe Sphinx is losing count of which code goes with which examples. My only guess would be that you need a blank line beneath `.. testcode::`?

mountain:
Hooray, it works. Thanks for your help, Brandon.

brandon-rhodes:
Thanks again for the contribution!

PR #308, opened 2019-12-15
Properly highlight Python in README

zdroid:


brandon-rhodes:
Interesting ­— I had thought that PyPI only supported plain RestructuredText? Is there documentation suggesting that this directive will correctly format there?

zdroid:
@brandon-rhodes I didn't align this with PyPI, since I'm not an user of it. The way I've given is how github's [linguist](https://github.com/github/linguist) parses code blocks. :computer:

brandon-rhodes:
The README is the file that gets posted here as the "project description":

https://pypi.org/project/skyfield/

It's difficult to test these days because once a new version is uploaded, its description can't be re-uploaded like they used to allow. So I think you'll want to find docs or just an existing example of the construct working, so we don't risk the next Skyfield release having wrongly formatted docs.

brandon-rhodes:
Okay, I found some docs:

https://packaging.python.org/guides/making-a-pypi-friendly-readme/#validating-restructuredtext-markup

If I misspell the directive deliberately as a test, I get:

```
$ twine check dist/*
...
    line 10: Error: Unknown directive type "code-blockm".
```

But if I switch back to `code-block` it says:

```
Checking dist/skyfield-1.15.tar.gz: PASSED, with warnings
  warning: `long_description_content_type` missing.  defaulting to `text/x-rst`.
```

So I'm going to go ahead and accept, and we'll see what happens when I next release!

zdroid:
It works for simpy, as you can see here:
https://pypi.org/project/simpy/
https://gitlab.com/team-simpy/simpy/raw/master/README.rst

zdroid:
Also it seems Markdown has been allowed aswell: https://packaging.python.org/guides/making-a-pypi-friendly-readme/

brandon-rhodes:
Great, thanks for the link to the SymPy example! I'll watch for the highlighting to change the next time Skyfield is released.

Issue #54, opened 2015-04-18
Improve Topos() behavior when caller swaps longitude and latitude


brandon-rhodes:
If the caller types

```
Topos('0 W', '90 N')
```

then the result at the moment is the only slightly informative

```
ValueError: your latitude string '0 W' does not end with either 'N' or 'S'
```

Instead, we should detect that the `W` is in fact one of the two possible endings that can be put after a longitude, and recommend that the user offer the attributes in reverse order. Better yet: instead of just saying “put them in reverse order,” we should print a message with the repr's of the two strings that the user has offered, so that (if they want) they can cut and paste from the error message directly back into their code, telling them that they can type

```
Topos('90 N', '0 W')
```

mysteriousham73:
Sorry for some of the confusion on this, first timer here trying to figure out what I'm doing!

brandon-rhodes:
After long experience with the `Topos` class, Skyfield is moving to a different system: users will ask an object representing an explicit datum like WGS84 to interpret the latitude and longitude angles. This will eliminate obvious user questions and confusion around which datum and which ellipsoid is being used to interpret lat/lon. And while making the shift, I have simplified the API to remove strings; instead, floating point degrees is what always must be specified. But to keep things readable, I now provide constants like `E` and `W` to supply the correct signs:
```
    boston = earth + wgs84.latlon(42.3583 * N, 71.0603 * W, elevation_m=43)
```
As the old string-based `Topos` class will no longer be documented (though it will continue to exist forever to keep old code working), let’s not plan on any further complications or enhancements to its API.

Users will hopefully supply latitude and longitude in the right order to the new `latlon()` method because the method has them in exactly that order — we’ll see if it works or not!

PR #329, opened 2020-01-29
Fix for #54

mysteriousham73:
This is my first contrib to any github project so please forgive any error in etiquette or process.

In looking at 54, I did choose to ignore the requirement to warn the user if they've reversed their N/S and E/W as there's enough information to just do the right thing.

I do get that you might want to warn them, and also you'll note my error checking really isn't totally robust.  I don't have a good feel for your style when it comes to the overhead of error checking so I just did what I would have done, with the knowledge that at least the technical solution is there if someone wants to add the error checking on top of it.

As a side note, I'm using skyfield in a standalone satellite tracking application and have enjoyed it enough that I wanted to contribute back, hence this PR.  Very pumped at the upcoming almanac stuff, I was in the process of implementing such a thing outside of skyfield when I saw the PRs.  

Finally, thanks for sharing skyfield, it's great!

### DIFF
--- a/skyfield/toposlib.py
+++ b/skyfield/toposlib.py
@@ -3,7 +3,7 @@ from numpy import einsum
 from .constants import ASEC2RAD, tau
 from .earthlib import terra
 from .functions import rot_x, rot_y, rot_z
-from .units import Distance, Angle, _interpret_ltude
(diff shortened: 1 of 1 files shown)


mysteriousham73:
Travis build didn't pass

PR #331, opened 2020-01-29
Fix for issue 54

mysteriousham73:
OK I figured out how to get Travis builds going in my repo, so this does pass now!  Thanks again for skyfield!

### DIFF
--- a/skyfield/toposlib.py
+++ b/skyfield/toposlib.py
@@ -70,7 +70,7 @@ class Topos(VectorFunction):
         self.x = x
         self.y = y
 
-        self.R_lat = rot_y(latitude.radians)[::-1]
+        self.R_lat = rot_y(self.latitude.radians)[::-1]
 
         self.target = object()  # TODO: make this more interesting
         self.target_name = '{0} N {1} E'.format(self.latitude, self.longitude)
--- a/skyfield/toposlib.py
+++ b/skyfield/toposlib.py
@@ -3,7 +3,7 @@ from numpy import einsum
 from .constants import ASEC2RAD, tau
 from .earthlib import terra
 from .functions import rot_x, rot_y, rot_z
-from .units import Distance, Angle, _interpret_ltude
+from .units import Distance, Angle, _interpret_ltude, _unsexagesimalize
 from .vectorlib import VectorFunction
 
 
@@ -28,26 +28,44 @@ class Topos(VectorFunction):
     def __init__(self, latitude=None, longitude=None, latitude_degrees=None,
                  longitude_degrees=None, elevation_m=0.0, x=0.0, y=0.0):
 
+        lookup = {'N': (1.0, 'latitude'), 'S': (-1.0, 'latitude'), 'E': (1.0, 'longitude'), 'W': (-1.0, 'longitude')}
+
+        """
+        The following code makes the assumption that:
+          1.  If latitude contains a string of the type '<float>[NS]' then it is a latitude and longitude will
+              contain a longitude of the type '<float>[EW]'.
+          2.  If latitude contains a string of the type '<float>[EW]' then reverse of the above is true.
+          3.  Whatever type is contained in latitude, same type will be contained in longitude.
+
+        As an addition, the elifs could test both latitude and longitude for type as opposed to just latitude, then
+        raise a generic error.
+        """
+
         if latitude_degrees is not None:
-            latitude = Angle(degrees=latitude_degrees)
-        elif isinstance(latitude, (str, float, tuple)):
-            latitude = _interpret_ltude(latitude, 'latitude', 'N', 'S')
-        elif not isinstance(latitude, Angle):
+            self.latitude = Angle(degrees=latitude_degrees)
+            self.longitude = Angle(degrees=longitude_degrees)
+
+        elif isinstance(latitude, str):
+            if lookup[latitude[-1]][1] == 'latitude':
+                self.latitude = Angle(degrees=float(latitude[:-1].strip()) * lookup[latitude[-1]][0])
+                self.longitude = Angle(degrees=float(longitude[:-1].strip()) * lookup[longitude[-1]][0])
+            elif lookup[latitude[-1]][1] == 'longitude':
+                self.latitude = Angle(degrees=float(longitude[:-1].strip()) * lookup[longitude[-1]][0])
+                self.longitude = Angle(degrees=float(latitude[:-1].strip()) * lookup[latitude[-1]][0])
+
+        elif isinstance(latitude, (float, tuple)):
+            self.latitude = Angle(degrees=_unsexagesimalize(latitude))
+            self.longitude = Angle(degrees=_unsexagesimalize(longitude))
+
+        elif isinstance(latitude, Angle):
+            self.latitude = latitude
+            self.longitude = longitude
+
+        else:
             raise TypeError('please provide either latitude_degrees=<float>'
                             ' or latitude=<skyfield.units.Angle object>'
                             ' with north being positive')
 
-        if longitude_degrees is not None:
-            longitude = Angle(degrees=longitude_degrees)
-        elif isinstance(longitude, (str, float, tuple)):
-            longitude = _interpret_ltude(longitude, 'longitude', 'E', 'W')
-        elif not isinstance(longitude, Angle):
-            raise TypeError('please provide either longitude_degrees=<float>'
-                            ' or longitude=<skyfield.units.Angle object>'
-                            ' with east being positive')
-
-        self.latitude = latitude
-        self.longitude = longitude
         self.elevation = Distance(m=elevation_m)
         self.x = x
         self.y = y


mysteriousham73:
With this PR I didn't do what you asked with the error message, rather I made it work even if the user swaps latitude and longitude since they're easy to discern by N/S and E/W.  If that's not your preferred solution I won't be offended but it seemed pretty straightforward and possibly less code to make it just work

brandon-rhodes:
> OK I figured out how to get Travis builds going in my repo, so this does pass now!

Welcome to GitHub and the adventure of contributing to open source!

If you'll review the GitHub instructions, I think you'll find that you can keep pushing to the same branch and repeatedly update a single pull request with improved code until you see that the tests pass. That would help you avoid creating several PR numbers, and would let you keep using the original one.

But, yes, starting Travis builds over in your own repo is even better (though it takes more work), since that lets you iterate as much as you want without anyone else seeing the errors, until you have the result green.

> With this PR I didn't do what you asked with the error message, rather I made it work even if the user swaps latitude and longitude since they're easy to discern by N/S and E/W. If that's not your preferred solution I won't be offended but it seemed pretty straightforward and possibly less code to make it just work

Thanks for thinking through this approach and working out how it looks in code! As I am reading through it:

* I like the idea of being able to also provide longitude and latitude as (degree, minute, second) tuples. That’s a great idea!
* However, as a long-term trend, I am starting to wonder if I did the right thing with all of these constructors with long lists of arguments of which only one can be useful. It leads to (a) lots of unused arguments (which affects speed slightly), (b) arguments that get ignored if some other argument is used (so if you provide longitude two different ways then it uses one and silently ignores the other), and (c) big `if` statements inside of every constructor (which is a cost both in speed, slightly, and also in complexity). So my approach seems to be a problem both for the code and for the users.
* So it might be better, rather than adding the (degree, minute, second) tuple as yet another way to provide longitude and latitude here, that it would be better to add a `Topos.from_dms(longitude, latitude)` constructor method that always expects its arguments to be (d,m,s) tuples. It would not only be able to run without any `if` statements inside at all, but would actually be readable ­— if someone reading code later thought "what are these three numbers", the `dms` might make them think of "degrees, minutes, and seconds". What do you think of that idea?
* Returning to your code: I worry that a user might try providing latitude one way and longitude another way. Your code does not seem to allow one of them to be a float and the other a string, but assumes without checking that they would both be provided the same way?
* But, we should notice, your idea here that "both arguments should always be provided exactly the same way" is similar to my idea of a `from_dms()` constructor: both approaches share in common the idea that arguments should be symmetrical; they just arrive there in different ways.
* You ask a very good question about argument order. My guess is that we should not allow re-ordering. Let's see what reasons I can come up with. (a) Allowing them in the other order would contradict what users see in their editor or IDE. They would see `'77.0669 W', '38.9215 N'` and, if they are still learning Skyfield and are not sure what those numbers are, they might highlight the function so their IDE shows them the names of the two parameters. They would be told the first one is `latitude=` which I think would be confusing. (b) If they use keyword arguments, the result is even worse, as the call reads `latitude='77.0669 W', longitude='38.9215 N'` which will be very hard on folks who are new to the concepts. (c) Really one of the most fragile parts of Skyfield is that return values only wind up assigned to the correct names because of consistent order. If someone types `dec, ra, distance = p.radec()`, there is nothing that will catch their error. Python does no checking of the names to which return values are assigned. Our only hope is that people always use the order "ra, dec", and the order "lat, lon". And my worry is that if we allow the order to be mixed up when they are input, then we weaken the consistency of their whole codebase, because the values are now appearing in different order in different places, whereas readability is improved if they are always the same.

All of those reasons, I think, would argue for taking the approach recommended in the TODO file. Do you have any objections that we should consider?

I would be interested to see how the code comes out if detection, rather than accepting the swapped arguments, becomes the goal. And, as stated above, I would also be happy to see a `from_dms()` class method to see what it would look like.

Thanks again for deciding to jump in and try to contribute! Hopefully with a slight change of direction, several of your lines of code will soon wind up becoming part of Skyfield!

mysteriousham73:
Thanks for the comments.  A few things:

1.  Regarding the complicated constructors, I do agree.  It's tempting to try and do it all but then there's also that line of where the module ends and the user code begins.  For what skyfield seems to be good at (using state-of-the-art techniques to do astronomical calculations), it seems like the ideal approach in situations like these is to have one constructor that takes the most basic input and let the user deal with any other formats.  I would argue this would be +/- decimal degrees in a float.

That said, I am sure at this point there's a concern about backwards compatibility so that horse is out of the barn and down the road.

On the point of backward compatibility, I didn't actually add the tuple DMS handling, it seemed to be in there already.  I was concerned about breaking backward compatibility so I kept it in, perhaps the way my code is organized made it stand out a little more.

I was also concerned about performance overhead which is why I decided to take the 'trust the user' approach here.  We can sanitize every input to every constructor and method, or do the opposite, or somewhere in between.  I don't have a feel for what level of paranoia you are comfortable with so I just kind of shot somewhere in the middle by type checking and leaving the other to trust.

So given that Python doesn't allow different constructors based on type, I think there are two approaches that might make sense:

1.  Keep the current Topos object and the current constructor.  Extend my solution to do further sanity checking for lat/long, advise the user per the todo, and leave it at that.  That will keep with your original intent but will be by far the heaviest implementation in terms of sanity checking.  This could possibly (?) be simplified if you're confident nobody is using tuples currently (again they were already in there, but perhaps not documented) by removing that handling.

2.  Deprecate the current Topos object and code a new one that takes only decimal degrees as inputs.  Sanity check those only to be sure they're within appropriate numeric bounds.  Minimal performance overhead and much less complicated, but breaks backwards compatibility. 

3.  Extend the current constructor to look something like this:

`def __init__(self, latitude=None, longitude=None, latitude_degrees=None,
                 longitude_degrees=None, elevation_m=0.0, x=0.0, y=0.0, latitude_tuple=None, longitude_tuple=None, latitude_string=None, longitude_string=None):`

In this implementation we would go full paranoid on the 'latitude' and 'longitude' named parameters with all the sanity checking and warnings.  However, if the user passes to the degree, tuple, or string named params the values get minimal sanity checking and get assigned straight to self.latitude and self.longitude.

This would keep backward compatibility if a user decides to use the original constructor style, albeit at some performance cost which it seems like you were willing to accept based on the original todo.  However, it would provide a fast lane should they decide to use the other params.  The downside to all this is further complicating the constructor with additional parameters.

So that's where I'm at on it, there's no perfect solution to this one due to the need for backward compatibility.  If you're inclined to break it some this could definitely be cleaned up quite a bit but I don't know what your installed base looks like.  It's worth noting that the docs do use the string format in their examples so that one at least is probably harder to get rid of since a lot of folks probably went that route.

brandon-rhodes:
> … it seems like the ideal approach in situations like these is to have one constructor that takes the most basic input and let the user deal with any other formats. I would argue this would be +/- decimal degrees in a float.

I agree, but with a caveat: the fundamental value, I think?, would be an Angle object providing the degrees. The reason I say that is that the Topos object stores them as Angle objects, and if the caller already has Angles, it makes sense to let them pass in the Angle objects instead of passing their bare degrees and forcing the `__init__()` method to re-create another Angle.

So accepting an Angle object handles all cases, whereas accepting only a float builds a redundant object in some cases.

(And I guess there's also an argument that radians are what Skyfield really cares about because that's what the underlying trig functions want?)

> That said, I am sure at this point there's a concern about backwards compatibility so that horse is out of the barn and down the road.

Yes! The current big constructor will stay.

> On the point of backward compatibility, I didn't actually add the tuple DMS handling, it seemed to be in there already.

Oh — look at that, it is there already! The fact that this diff imports `_unsexagesimalize` here for the first time made me for a moment think that the feature must have been missing.

> Like you, I was also concerned about performance overhead … I don't have a feel for what level of paranoia you are comfortable …

That's an excellent question! The answer happens to be that, when it comes to backwards compatibility for documented interfaces, I am utterly paranoid. So if the code previously allowed "mixed" methods for specifying geographic coordinates, I think we have to continue to allow it.

> So given that Python doesn't allow different constructors based on type, I think there are three approaches that might make sense:
> 
> 1. Keep the current Topos object and the current constructor.  Extend my solution to do further sanity checking for lat/long, advise the user per the todo, and leave it at that.  That will keep with your original intent but will be by far the heaviest implementation for performance in terms of sanity checking.

I am not sure I yet see the performance hit. The logic could live inside of `_interpret_ltude()` in its `else:` clauses, where we already know the value is bad and we are going to have to raise an exception. In that case there is nothing to be lost in doing another string test or two and figuring out if a swapped value has been passed, I don't think?

> This could possibly (?) be simplified if you're confident nobody is using tuples currently…

I am not confident, and I think your argument is solid here: we have to leave the DMS tuple interpretation intact because it could already be in use. (Again, I only forgot that fact because the `import` statement threw me off.)

> 2. Deprecate the current Topos object and code a new one that takes only decimal degrees as inputs. …

I think that alternative constructors could solve the problem. (See below.)

> 3. Extend the current constructor to look something like this …:
> 
> `def __init__(self, latitude=None, longitude=None, latitude_degrees=None, longitude_degrees=None, elevation_m=0.0, x=0.0, y=0.0, latitude_tuple=None, longitude_tuple=None, latitude_string=None, longitude_string=None):`

I am not sure that has much advantage, since the same `if` statements still would have to be present; they would just be making slightly different tests, I think?

> P.S. I am admittedly ignoring the .from_dms() approach in lieu of #3 above as I believe you would still have the full overhead of the original init implementation being called? If that’s incorrect perhaps you can say in general terms how that would work?

Could we use an approach like this to avoid alternate constructors having to call a big ugly main `__init__()`?

https://stackoverflow.com/questions/2168964/python-creating-class-instance-without-calling-initializer

That way we could have three or four separate very simple constructors, and if we shifted the documentation to only ever mention those four, then hopefully use of the big ugly original one would wither away.

But, just to be clear, for this pull request, it will be enough to add the protection against swapped arguments! If you are interested in helping clean up the construtor story here, which would be great, let's do it next in another pull request, and have you start with just documentation for how they would look so we can agree on the design before you spend time writing up the implementations.

mysteriousham73:
So first, let me say that I really appreciate your kindness on this.  I recognize that you’re being patient with me on this and that’s pretty laudable considering that we’ve all easily figured out by now that I’m not a professional programmer.  In fact, the past couple weeks or so are my first visit back to Python in quite awhile, as evidenced by the fact that I was unaware of the appearance of a the “alternative constructors” you describe.  So thank you also for the education on that.

I do want to fix the constructor issue as I’m completely bothered by it now and want to see it through.  I will open another PR sometime soon here.  In the meantime I will finally take the bait properly that you set out for some newbie programmer and add a warning to _interpret_ltude :)

brandon-rhodes:
> … we’ve all easily figured out by now that I’m not a professional programmer. In fact, the past couple weeks or so are my first visit back to Python in quite awhile …

If it's any encouragement, the code you submitted looked clean enough that I made no guess about your experience level. In any case, welcome back to Python!

> I do want to fix the constructor issue as I’m completely bothered by it now and want to see it through. I will open another PR sometime soon here. In the meantime I will finally take the bait properly that you set out for some programmer like myself and add a warning to _interpret_ltude :)

For what it’s worth, my own approach towards problems is very often the route you’ve taken here. I step back, I think “but why isn’t it written _this_ way?”, and try rewriting some code in a different direction. Then I realize the problem, or see a test fail that tests a situation I did not even think about, and suddenly say _Oh okay that’s why it was done that way!_ Then I do the simpler thing instead.

brandon-rhodes:
@mysteriousham73 — Let me know whether you plan to make further progress on this pull request, or whether it should be closed for now. It would help clear the Skyfield dashboard a bit :) Thanks for any update you can provide!

brandon-rhodes:
As you are doubtless busy these last few months, I'm going to close this PR for now to keep my dashboard clear. Thanks for your interest in Skyfield, though, and feel free to comment here in the future if you want to continue the conversation!

PR #340, opened 2020-02-17
Adding tests for KeplerLib

JoshPaterson:
Sorry I wasn't able to respond earlier to your acceptance of #253, I'm very excited about it!

Since the last time I posted an update about `keplerlib.py` I've compared the results produced by the original FORTRAN function `prop2b` in the Spice library with `keplerlib.propagate` for a wide range of orbit types and times, and in the worst case scenario they differ in the 11th or 12th significant digit, so I feel pretty confident that `keplerlib.propagate()` is working properly.

One thing I want to mention is that I still don't think the `KeplerOrbit` object matches the data produced by Horizons, and I still don't think we should expect it to. To help explain why I'll give numbers to the steps the `KeplerOrbit` object uses to produce results so I can refer to them later:

1. Take a set of orbital elements and convert them into state vectors. This happens in `keplerlib.ele_to_vec()`, which is identical to the function `test_elementslib.ele_to_vec()` (it should probably live in one place and be imported to the other). It was originally written to help verify that `elementslib.py` was working properly.
2. Propagate the state vectors from step 1 to any other time. This happens in `keplerlib.propagate()`

The way `test_keplerlib.test_against_horizons` is currently written it's testing step one, but not step two. The set of orbital elements and state vectors from `horizons/ceres-orbital-elements` and `horizons/ceres-position` match each other exactly by definition because the elements are osculating elements. Because the time used to initialize the `KeplerOrbit` object is the same time given to the `KeplerOrbit._at()` method, the change in time is 0, so no propagation is happening; `keplerlib.propagate()` outputs the same vectors that it's given. In order to actually test step 2, the time given to the `KeplerOrbit._at()` method should be different from the one used to initialize the `KeplerOrbit` object. If you do this I think you'll find that the farther in time you propagate the orbit away from the time used for initialization, the greater it will differ from Horizons, due to Horizons not using 2 body propagation. I think the solution is to just rename the test to better describe what it does.

Currently the elements that are checked in the tests that make a round trip from elements to vectors and back are the ones that remain constant for Keplerian Orbits. These tests pass and they show that the propagated positions are on the correct orbit. What they don't show is that the positions are in the right place on the orbit. This PR adds a check of the mean anomaly which should verify that the propagated positions are in the right place in the orbits. The new check doesn't currently pass, but I think it ought to, and I'm working on figuring out why it doesn't.

### DIFF
--- a/skyfield/tests/test_keplerlib.py
+++ b/skyfield/tests/test_keplerlib.py
@@ -7,6 +7,8 @@ from skyfield.api import load
 import os
 
 seterr(all='raise')
+#I encountered an underflow in keplerlib.stumpff where c3 is calculated using power series. I think it should be ignored.
+seterr(under='ignore') 
 
 # Test against HORIZONS.
 
--- a/skyfield/tests/test_keplerlib.py
+++ b/skyfield/tests/test_keplerlib.py
@@ -54,6 +54,10 @@ def test_ele_to_vec():
 def _data_path(filename):
     return os.path.join(os.path.dirname(__file__), 'data', filename)
 
+times = linspace(-1e11, 1e11, 1001) # -3170 years to +3170 years, including 0 (in seconds)
+mu = 403503.2355022598
+dummy_time = load.timescale().utc(2018)
+
 def check_orbit(p, e, i, Om, w, v,
                 p_eps=None, e_eps=None, i_eps=None, Om_eps=None, w_eps=None, v_eps=None):
     pos0, vel0 = ele_to_vec(p, e, i, Om, w, v, mu)
--- a/skyfield/tests/test_keplerlib.py
+++ b/skyfield/tests/test_keplerlib.py
@@ -12,7 +12,7 @@ seterr(under='ignore')
 
 # Test against HORIZONS.
 
-def test_against_horizons():
+def test_ele_to_vec():
     # See the following files in the Skyfield repository:
     #
     # horizons/ceres-orbital-elements
--- a/skyfield/tests/test_keplerlib.py
+++ b/skyfield/tests/test_keplerlib.py
@@ -1,6 +1,6 @@
-from numpy import pi, seterr, linspace
+from numpy import pi, seterr, linspace, float_
 from skyfield.keplerlib import KeplerOrbit, propagate
-from skyfield.elementslib import OsculatingElements
+from skyfield.elementslib import OsculatingElements, mean_anomaly, eccentric_anomaly
 from skyfield.units import Angle, Distance, Velocity
 from skyfield.tests.test_elementslib import compare, ele_to_vec
 from skyfield.api import load
@@ -59,7 +59,7 @@ mu = 403503.2355022598
 dummy_time = load.timescale().utc(2018)
 
 def check_orbit(p, e, i, Om, w, v,
-                p_eps=None, e_eps=None, i_eps=None, Om_eps=None, w_eps=None, v_eps=None):
+                p_eps=None, e_eps=None, i_eps=None, Om_eps=None, w_eps=None, M_eps=None):
     pos0, vel0 = ele_to_vec(p, e, i, Om, w, v, mu)
 
     pos1, vel1 = propagate(pos0, vel0, 0, times, mu)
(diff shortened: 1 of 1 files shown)


JoshPaterson:
When I adjusted the tolerance for the mean anomaly tests, I found that the elliptical and parabolic orbits passed, but the circular and hyperbolic orbits didn't. I think this means that the problem is probably in `elementslib.py`, so I've removed those failing tests for now until I can figure out exactly what the problem is. I think for now they shouldn't be a reason for `test_keplerlib.py` to fail.

brandon-rhodes:
> Sorry I wasn't able to respond earlier to your acceptance of #253, I'm very excited about it!

It had waited for a while, so I was not expecting an immediate response. :)  I'm glad you've seen the notice, though!

> Since the last time I posted an update about `keplerlib.py` I've compared the results produced by the original FORTRAN function `prop2b` in the Spice library with `keplerlib.propagate` for a wide range of orbit types and times, and in the worst case scenario they differ in the 11th or 12th significant digit, so I feel pretty confident that `keplerlib.propagate()` is working properly.

Nice! So not all the way to the 15th or 16th digit, but pretty close agreement anyway. What is the physical meaning of the 12th digit — are we talking about fractions of a meter difference?

> … The way `test_keplerlib.test_against_horizons` is currently written it's testing step one, but not step two.

I hadn't seen any tests yet that asked "is step 1 performed correctly", so I thought it worth adding one. I always like having tests of individual steps, when I can, because it helps isolate problems later ("did I just break step 1, step 2, or both?" after changing something).

> The set of orbital elements and state vectors from `horizons/ceres-orbital-elements` and `horizons/ceres-position` match each other exactly by definition because the elements are osculating elements.

But quite a bit of code is still thereby being tested. I'm not unhappy with the test :) No objection to renaming it, though.
 
> The new check doesn't currently pass, but I think it ought to, and I'm working on figuring out why it doesn't.

Thanks for jumping back in to keep the code moving forward!

brandon-rhodes:
> When I adjusted the tolerance for the mean anomaly tests, I found that the elliptical and parabolic orbits passed, but the circular and hyperbolic orbits didn't. I think this means that the problem is probably in `elementslib.py`, so I've removed those failing tests for now until I can figure out exactly what the problem is. I think for now they shouldn't be a reason for `test_keplerlib.py` to fail.

Are circular orbits stable? I thought it was impossible to specify a circular orbit with Keplerian parameters because there's no periapsis to measure the mean anomaly against.

barrycarter:
Circular or even near-circular orbits are problematic. See https://astronomy.stackexchange.com/questions/28691/why-is-neptune-true-anomaly-decreasing/28704#28704 for details.

JoshPaterson:
> What is the physical meaning of the 12th digit — are we talking about fractions of a meter difference?

The orbits I was using for testing were all of a moon-like orbit, but with different eccentricities. The worst case scenario was a parabolic orbit being propagated 3000 years, in which case the 11th digit is 10s of meters out of about 9 au. Elliptical orbits were in general an order of magnitude better than parabolic orbits, and the difference was also proportional to time, so for an elliptical orbit propagated 30 years the difference was around a tenth of a millimeter.

> Are circular orbits stable? I thought it was impossible to specify a circular orbit with Keplerian parameters because there's no periapsis to measure the mean anomaly against.

That's a great point! I shouldn't be expecting mean anomaly to be useful for circular orbits. I'll still look into the hyperbolic case, but at this point I have no reason to think `keplerlib.py` isn't working correctly.

brandon-rhodes:
> I'll still look into the hyperbolic case, but at this point I have no reason to think keplerlib.py isn't working correctly.

Let me know when you're done looking, and I can start a review (or you can just click "request review" now in GitHub, I think?).

brandon-rhodes:
@JoshPaterson —

Do you happen to have handy any of the tests you did between the output of the new `propagate()` routine and the one in Spice that it's based on? Trying to propagate Hale-Bopp as a test case, I'm seeing what looks kind of like an orbit but has discontinuous leaps that create gaps during which different positions are given:

image

I produced that with:

```
import numpy as np
from skyfield.constants import AU_KM
from skyfield.keplerlib import KeplerOrbit, propagate
r = np.array([ 2.65881889e+08,  2.45099677e+08, -4.05820610e+09])
v = np.array([ 0.81512382, -0.98497339, -7.65759902])
t0 = 0
t1 = -22291134
gm = 132712440018.9
pos, vel = propagate(r, v, t0, t1, gm)
print(pos / AU_KM)

t1_array = np.linspace(-369359937*2, 0, 200)
x = []
y = []
for t1 in t1_array:
    print(t1)
    pos, vel = propagate(r, v, t0, t1, gm)
    x.append(pos[0])
    y.append(pos[1])

plt.plot(x, y)
plt.grid(True)
plt.savefig("test.png")
```

Do you have any guess about where in the routine the discontinuities are appearing? Also, let me know if it looks like I'm calling it correctly. (Here, I'm just trying for single values. After I figure out how to make something that looks like an orbit by calling it with scalars, I'll get more brave and try passing it a vector.)

Thanks for any advice you might have!

JoshPaterson:
I swapped in the fortran version of `propagate()` and this is the plot I got:

test_fortran

Is that what you expected? Either way that's definitely an issue with `keplerlib.propagate()`. I'll take a look in the debugger to see where they differ.

brandon-rhodes:
> I swapped in the fortran version of propagate()…

Great idea! Yes, I expected more of a smooth curve like that (though I'd kind of expected it to be shorter, more the shape of the little arc up at the top of my plot). Thanks for taking a look, and I look forward to hearing what you find!

JoshPaterson:
I realized I mistakenly had `dt = t1 - t0`, but when I corrected that to `dt = t1 - t0` the FORTRAN version actually looks like this:

test_both

brandon-rhodes:
Ah, good, that is more similar to what I had expected! If you'll switch the order in which you graph them, and let the Python curve be painted on top of the FORTRAN one, it will be a little bit clearer where the Python curve agrees with it (since right now at the places of agreement the Python one is hidden). Or I guess you could adjust the FORTRAN curve's opacity?

In any case, I'm happy we have visual confirmation that Python sometimes agrees with but sometimes diverges from the FORTRAN number. Good luck as you track down the places in the code where they differ (do modulo operations come in anywhere?), and let me know if you have any questions I can answer!

JoshPaterson:
Good news! I figured out the problem, and just made a PR to fix it. Here is the plot now:

test_both

JoshPaterson:
Also, thanks for working on this PR! I haven't been able to work on it as much as I would like to lately, so I appreciate you doing that.

brandon-rhodes:
@JoshPaterson Could you share your code that swaps in the FORTRAN version of the function? I'd like to try taking the difference between the two routines to see how large the difference is over those same years you're graphing. Thanks!

JoshPaterson:
I made a gist with the fortran source code, the compiled code, and a python wrapper. I also included the unvectorized version of `keplerlib.propagate()` because it's a lot more readable than the vectorized version, so if you're wondering why the vectorized `skyfield.keplerlib` version is doing something it can help to look at the unvectorized version.

https://gist.github.com/JoshPaterson/5e203f1bc69db38a5687be8b73e64779

I tried to make the readme as clear as possible, but if you have any questions let me know :)

brandon-rhodes:
This will help me as I work on comets this weekend. Thanks so much!

brandon-rhodes:
@JoshPaterson — Did you work out whether this PR/branch had any work you needed to salvage, or whether it can safely be closed? Just asking because it would clean up my dashboard a bit; no rush!

JoshPaterson:
I haven't yet, but I think we could just close this one and I'll open a new one if necessary.

PR #342, opened 2020-02-17
Fix math error

JoshPaterson:
I found this math error when I tried changing t here:

https://github.com/skyfielders/python-skyfield/blob/e58b514e98128e08b336f960ea12502bf97bbcbd/skyfield/tests/test_keplerlib.py#L41

and I couldn't get the test to fail.

### DIFF
--- a/skyfield/keplerlib.py
+++ b/skyfield/keplerlib.py
@@ -54,7 +54,7 @@ class KeplerOrbit(VectorFunction):
         if mu_km_s:
             self._mu_km_s = mu_km_s
         elif mu_au_d:
-            self._mu_km_s =  mu_au_d / AU_KM**3 * DAY_S**2
+            self._mu_km_s =  mu_au_d * AU_KM**3 / DAY_S**2
 
         self.position_at_epoch = position
         self.velocity_at_epoch = velocity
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thanks for the update!

Longer-term (like, maybe later this month) I would like to simplify to only having one of these two ways of providing the value, to keep the signature simpler. Generally I have been having Skyfield always use au and days internally, so my preference will probably be to keep that one.

JoshPaterson:
I think switching to au^3/d^2 would be great. 

The only reason I used km^3/s^2 was because that's what I used internally in `elementslib.py`, and the only reason I did that was because that was the unit used in the file I copied and pasted from Horizons into `data/gravitational_parameters.py`. I've since realized it would have been easy to use a dictionary comprehension to make a version of `data.gravitational_parameters.GM_dict` that uses au^3/d^2 to match the working units of the rest of skyfield, but I didn't think of it at the time!

barrycarter:
Interestingly(ish), au^3/d^2 is what NASA uses internally to compute the developmental ephemeredes: http://ilrs.gsfc.nasa.gov/docs/2014/196C.pdf explains more and the raw data is at: ftp://ssd.jpl.nasa.gov/pub/eph/planets/ascii/de431/header.431_572

Less interestingly, I make an attempt to explain the data myself at: https://github.com/barrycarter/bcapps/tree/master/ASTRO/README

brandon-rhodes:
Going forward, we might want to think about having users import official NASA files for solar system masses instead of carrying them inline, since they are values that get revised and updated. Using the parser I already wrote for [planetary reference frames](https://rhodesmill.org/skyfield/planetary.html) we should be able to import a file like:

https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/gm_de431.tpc

(Which, I see, uses the km-s rather than the au-d that would more perfectly match Skyfield!)

PR #352, opened 2020-04-02
Fix power series

JoshPaterson:
Fixes the problem discussed in #340

Previously every term in the power series for the stumpff function was summed, but this commit fixes it so that the terms are alternately added and subtracted.


brandon-rhodes:
Thanks! I'm glad the problem didn't tangle you up in needing to change several places in the code, but was happily confined to one spot that you could fix. Good going!

PR #356, opened 2020-04-13
Throw exception if calculation time is out of range

### LINKED ISSUE #319: Add support for jplephem's new OutOfRangeError exception
A new exception has been introduced in upstream jplephem library that allows better catching of out-of-range timesstamps (brandon-rhodes/python-jplephem#36).

This exception should be caught by Skyfield to raise a new one, which includes `Time` instances with the minimum and maximum datetimes supported by jplephem.

See the discussion in upstream PR.

Deuchnord:
Add support for the `OutOfRangeError` exception thrown by JplEphem in brandon-rhodes/python-jplephem#36.

Basically, we catch the raised exception on JplEphem's side and re-throw a Skyfield-specific one with the times converted in `Time` objects.

Fix #319

### DIFF
--- a/skyfield/jpllib.py
+++ b/skyfield/jpllib.py
@@ -3,10 +3,12 @@
 import os
 from collections import defaultdict
 
+from jplephem.exceptions import OutOfRangeError
 from jplephem.spk import SPK
 from jplephem.names import target_name_pairs, target_names as _names
 
 from .constants import AU_KM, DAY_S
+from .errors import OutOfRangeTimeError
 from .timelib import calendar_date
 from .vectorlib import VectorFunction, VectorSum
 
@@ -204,8 +206,21 @@ class SPICESegment(VectorFunction):
 
 class ChebyshevPosition(SPICESegment):
     def _at(self, t):
-        position, velocity = self.spk_segment.compute_and_differentiate(t.tdb)
-        return position / AU_KM, velocity / AU_KM, None, None
+        try:
+            position, velocity = self.spk_segment.compute_and_differentiate(t.tdb)
+            return position / AU_KM, velocity / AU_KM, None, None
+        except OutOfRangeError as e:
+            start_time = t.ts.tdb(jd=self.spk_segment.start_jd)
+            end_time = t.ts.tdb(jd=self.spk_segment.end_jd)
+            out_of_range_times = []
+
+            for i, time in enumerate(t):
+                if e.out_of_range_times[i]:
+                    out_of_range_times.append(time)
+
+            raise OutOfRangeTimeError('out of range times: must be between %s and %s' % (start_time.utc_iso(),
+                                                                                         end_time.utc_iso()),
+                                      start_time, end_time, out_of_range_times)
 
 
 class ChebyshevPositionVelocity(SPICESegment):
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
Thanks, hopefully I'll have a chance to take a look later this week!

skyfield/jpllib.py
```diff
@@ -204,8 +206,21 @@ def __repr__(self):
 
 class ChebyshevPosition(SPICESegment):
     def _at(self, t):
-        position, velocity = self.spk_segment.compute_and_differentiate(t.tdb)
-        return position / AU_KM, velocity / AU_KM, None, None
+        try:
+            position, velocity = self.spk_segment.compute_and_differentiate(t.tdb)
+            return position / AU_KM, velocity / AU_KM, None, None
```
brandon-rhodes, line 211: It's a general principle in programming to move as many lines outside of each `try:` as possible to keep software as easy to reason about as possible, so let's move this `return` down outside the `try…except`. That way it's visually more obvious to later readers that there's nothing that could go wrong with the division that the `except` is trying to catch.

skyfield/jpllib.py
```diff
@@ -204,8 +206,21 @@ def __repr__(self):
...
+        except OutOfRangeError as e:
+            start_time = t.ts.tdb(jd=self.spk_segment.start_jd)
+            end_time = t.ts.tdb(jd=self.spk_segment.end_jd)
+            out_of_range_times = []
+
+            for i, time in enumerate(t):
```
brandon-rhodes, line 217: I agree that we want to give users information about which times were the problem, but I'm not sure an array of separate `Time` objects, one per value, will be very efficient — if they passed us an array of several thousand times, most of which were out of range, this loop would take a bit of time to create thousands of `Time` objects.

It would be simplest to offer users the bare Boolean mask `e.out_of_range_times` (it's a list of Trues and Falses, if I'm reading right?) as an attribute. Then they would be free to process that information as they wish, and if they didn't need separate `Time` objects created for every bad value, they'd be saved the expense.

Plus: they would know immediately which array elements were bad. If we simply return the values of the bad times, then they'd have to search their time array manually for each value to know where they are.

skyfield/jpllib.py
```diff
@@ -204,8 +206,21 @@ def __repr__(self):
...
+                if e.out_of_range_times[i]:
+                    out_of_range_times.append(time)
+
+            raise OutOfRangeTimeError('out of range times: must be between %s and %s' % (start_time.utc_iso(),
```
brandon-rhodes, line 221: Perfect! The big question they'll have, "what times are supported by the ephemeris?", is answered right in the text of the exception. This will be an improvement for users!

Deuchnord:
Hi @brandon-rhodes, sorry for the late, didn't have much time this week to fix the review.

> Also: could you add a unit test for the feature? You'll want to create a new file test_jpllib.py since it's a test specific to this suite.

I'm not sure how I should proceed for that… at first sight, looks like I should mock the JplLib library, but I don't know the tests framework you're using… do you have any doc for it?

brandon-rhodes:
Thanks for contributing this! The feature should appear in the next release of Skyfield.

PR #363, opened 2020-04-24
Update urlopen call to not use deprecated cafile keyword

eteq:
Skyfield's `iokit.py` is currently using the `cafile` keyword in `urlopen`, which is deprecated from Python 3.6 on (see the deprecation note at the very end of the `urlopen` section here: https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen). This PR updates the call to `urlopen` to use what I think to be the intended replacement.

While in principle a minor issue, it turns out to lead very indirectly to some failing tests downstream - see astropy/astropy#10200 for more detail on what lead me down this path... But regardless of the downstream effect, this would seem to be more what the current python standard library intends!

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -42,6 +42,7 @@ try:
 except:
     from urlparse import urlparse
     from urllib2 import urlopen
+from ssl import create_default_context
 
 # If we are running under the built-in IDLE development environment, we
 # cannot use '\r' to keep repainting the current line as a progress bar:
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Alas! It's a mere DeprecationWarning, but since it's occurring in a test suite, it's upgraded to a fatal error. Your diff will break older Pythons, but I'm going to go ahead and let it land, then fix the breakage in CI with an `if` statement, and then release a new version immediately.

Thanks VERY much for letting me know so quickly — the Python 3.6 that I'm using hadn't raised any warnings!

brandon-rhodes:
@eteq — I've just released Skyfield 1.20, which should fix this for all Python versions. Thanks again!

eteq:
Thanks for the very fast turn around on this, @brandon-rhodes!

PR #365, opened 2020-04-26
Fix typo in the EphemerisRangeError constructor

Deuchnord:
Fixing a small mistake introduced in commit c2181134d70a2a8ba104ba0db5b39b6cb633263a, in the constructor of the new `EphemerisRangeError`.


brandon-rhodes:
Alas, a classic cut and paste error! How did I do it twice, though? :(  Thanks VERY much for reading the next diff and double-checking me!

PR #366, opened 2020-04-27
Added timescale passthrough for EarthSatellite instantiation via Loader.tle_file()

bwburnsides:
Currently the `EarthSatellite` class is instantiated with a `ts` `kwarg`, but if you instantiate via the `Loader.parse_tle_file()` method, `ts` specification is lost. I added a simple pass through in the form of a `ts` `kwarg` for `Loader.tle_file()` and `Loader.parse_tle_file()` so that the `EarthSatellite` instances created by the `parse_tle_file()` generator can have specific timescales.

### DIFF
--- a/skyfield/iokit.py
+++ b/skyfield/iokit.py
@@ -246,7 +246,7 @@ class Loader(object):
                     d[name] = sat
         return d
 
-    def tle_file(self, url, reload=False, filename=None):
+    def tle_file(self, url, reload=False, filename=None, ts=None):
         """Load and parse a TLE file, returning a list of Earth satellites.
 
         Given a URL or local path to an ASCII text file, this loads a
@@ -259,7 +259,7 @@ class Loader(object):
 
         """
         with self.open(url, reload=reload, filename=filename) as f:
-            return list(parse_tle_file(f))
+            return list(parse_tle_file(f, ts=ts))
 
     def open(self, url, mode='rb', reload=False, filename=None):
         """Open a file, downloading it first if it does not yet exist.
@@ -532,7 +532,7 @@ def parse_tle(fileobj):
         b0 = b1
         b1 = b2
 
-def parse_tle_file(lines):
+def parse_tle_file(lines, ts=None):
     """Parse TLE satellite elements, yielding a sequence of satellites.
 
     Given a sequence ``lines`` of byte strings that contain TLE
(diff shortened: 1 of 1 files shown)


skyfield/iokit.py
```diff
@@ -561,7 +561,7 @@ def parse_tle_file(lines):
 
             line1 = b1.decode('ascii')
             line2 = b2.decode('ascii')
-            yield EarthSatellite(line1, line2, name)
+            yield EarthSatellite(line1, line2, name=name, ts=ts)
```
brandon-rhodes, line 564: And the same here: positional args will be more efficient.

skyfield/iokit.py
```diff
@@ -259,7 +259,7 @@ def tle_file(self, url, reload=False, filename=None):
 
         """
         with self.open(url, reload=reload, filename=filename) as f:
-            return list(parse_tle_file(f))
+            return list(parse_tle_file(f, ts=ts))
```
brandon-rhodes, line 262: Let's switch this to just `(f, ts)`. Adding a keyword argument when a positional would work will make this call more expensive for Python, and the value of the keyword for documentation is not that high here, since the routine being called is right in the same file.

bwburnsides:
Hi Brandon, thanks for getting back to me on this. I will change the `kwargs` to positional arguments. The reason I used `kwargs` in the fist place is to prevent existing uses of these methods from failing.

Also, what form of testing are you looking for? I am new to unit testing and am so far only familiar with `pytest`.

brandon-rhodes:
> The reason I used kwargs in the first place is to prevent existing uses of these methods from failing.

Yes, and that was a good choice! To clarify: I very much want the parameters to remain optional so existing calls keep working. It's only in the private invocations we make, which don't affect other code, that I'm asking for a pivot back to a positional calling style.

> Also, what form of testing are you looking for? I am new to unit testing and am so far only familiar with pytest.

Come to think of it, it will probably be faster for me to add it — I'll just do it in the documentation that mentions the feature. So you can skip the testing part and I'll add it once it lands.

bwburnsides:
Okay, sorry for the confusion - I thought this was going to be a bigger change than it was. I didn't know that if you supplied excess positional arguments, that they would automatically fill the `kwargs`. Learn something new everyday; everything should be good to go now.

brandon-rhodes:
Thanks, you can expect to see this feature appear officially in the next version of Skyfield!

PR #368, opened 2020-04-29
added xp/yp passthrough in sgp4lib.ITRF_position_velocity_error

bwburnsides:
I added `0.0`-defaulted polar motion pass-through `kwargs` to `sgp4lib.ITRF_position_velocity_error()`. Currently in my project I wish to convert SGP4 propagated TEME coords to ITRF coords and wish to specify polar motion coefficients. Currently, `ITRF_position_velocity_error()` calls `TEME_to_ITRF()` without specifying the coefficients, so they are always kept as `0.0`. To get around this, I have had to manually call the private `_position_and_velocity_TEME_km()` and then `TEME_to_ITRF()`. This is not a clean solution, especially as `TEME_to_ITRF()` already accepts polar motion coefficients as arguments.

### DIFF
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -138,7 +138,7 @@ class EarthSatellite(VectorFunction):
             message = SGP4_ERRORS[error] if error else None
             return array(position), array(velocity), message
 
-    def ITRF_position_velocity_error(self, t):
+    def ITRF_position_velocity_error(self, t, xp=0.0, yp=0.0):
         """Return the ITRF position, velocity, and error at time `t`.
 
         The position is an x,y,z vector measured in au, the velocity is
@@ -150,7 +150,7 @@ class EarthSatellite(VectorFunction):
         rTEME /= AU_KM
         vTEME /= AU_KM
         vTEME *= DAY_S
-        rITRF, vITRF = TEME_to_ITRF(t.ut1, rTEME, vTEME)
+        rITRF, vITRF = TEME_to_ITRF(t.ut1, rTEME, vTEME, xp, yp)
         return rITRF, vITRF, error
 
     def _at(self, t):


brandon-rhodes:
> This is not a clean solution, especially as TEME_to_ITRF() already accepts polar motion coefficients as arguments.

Interesting! It may take me a few days to form my thoughts on this one, as polar motion coefficients aren't something I've touched in a while.

May I ask what your application is that has TLEs so sensitive that polar motion coefficients arise as a useful correction? Having a larger picture of what's being computed and at what accuracy would help me as I think about how to better support polar motion.

bwburnsides:
We are interested in creating high-fidelity link budgets between specific satellites and a moving target, as well as moderate-fidelity (as much as TLEs may provide) collision avoidance metrics. My current solution (calling the private method directly and then TEME_to_ITRF explicitly) works well, so this is more a way to follow best practices and increase usability rather than a change to functionality.

brandon-rhodes:
Happily, I think that you can neglect polar motion for your application. TLEs are only advertised as accurate to within a few kilometers. We can take, say, the ISS moving overhead and jiggle the predicted position 1km in each direction to model the uncertainly and learn its effect on the azimuth and altitude at which we would point an antenna:

```
from skyfield.api import Distance, EarthSatellite, Topos, load
from skyfield.functions import length_of

ts = load.timescale(builtin=True)
sat = EarthSatellite(
    '1 25544U 98067A   20120.54828704  .00001508  00000-0  35122-4 0  9994',
    '2 25544  51.6452 229.1987 0001494 205.5139 329.7843 15.49330212224471',
)
topos = Topos('40.8939 N', '83.8917 W')

t = ts.utc(2020, 4, 29, 9, 33)

alts = []
azs = []

import itertools

s = (sat - topos).at(t)

offsets = [-1, +1]
for dx, dy, dz in itertools.product(offsets, offsets, offsets):
    old = s.position
    s.position = Distance(km=old.km + [dx, dy, dz])
    alt, az, distance = s.altaz()
    alts.append(alt.degrees)
    azs.append(az.degrees)
    s.position = old

print('Altitude:', min(alts), '-', max(alts), 'delta', max(alts) - min(alts))
print('Azimuth:', min(azs), '-', max(azs), 'delta', max(azs) - min(azs))
print('size of polar motion "y" in degrees: %f' % (0.331 / 3600.0))
```

Output:

```
Altitude: 44.492668354090064 - 44.82178094438337 delta 0.3291125902933061
Azimuth: 268.1974534768886 - 268.6131686054818 delta 0.4157151285932059
size of polar motion "y" in degrees: 0.000092
```

If I've written this script correctly, it shows that the inherit inaccuracy of a TLE is going to produce angular errors for the ISS of ⅓ to nearly ½ a degree. Within that error, the 0.000092 degree effect of polar motion will be swamped. An antenna capable of correcting a ⅓ degree error and zeroing in on the satellite won't be upset by an additional 0.000092 degrees of error.

Or is your target not in low earth orbit, but an orbit for which the angular position will be known to greater accuracy?

brandon-rhodes:
(Aside: I have no idea why I dragged `itertools` into that example. Here's a substitute loop that's much easier to read:)
```
for dx in offsets:
    for dy in offsets:
        for dz in offsets:
            old = s.position
            s.position = Distance(km=old.km + [dx, dy, dz])
            alt, az, distance = s.altaz()
            alts.append(alt.degrees)
            azs.append(az.degrees)
            s.position = old
```

brandon-rhodes:
I am going to close this for now. I probably should not have included polar motion in `TEME_to_ITRF()` as it's not a realistic adjustment to be making to a real-world conversion of satellite coordinates; I only added xp and yp there so that the test of Vallado’s Appendix C could succeed to high precision, but I really instead should have just done those two rotations right in the test, and not included them as public arguments to the routine.

I can't remove xp and yp from `TEME_to_ITRF()` now, alas, without potentially breaking someone's code, but I am inclined to not let them spread further. Thanks for pushing me to think that through, though!

bwburnsides:
I'm not sure how these messages from April flew under my radar for so long. Apologies! I agree that polar motion is negligible for nearly any application TLEs would be used in. Makes sense to limit they're use in the public API. Thanks!

Issue #362, opened 2020-04-21
Feature Request: Add eclipse/illumination times or something similar to the EarthSatellite object


jacoffey3:
Basically I'm looking for an understanding of when the satellite is illuminated by the sun, and when it is in eclipse. There are a few posts about this I've found in other places: 

https://space.stackexchange.com/questions/37713/how-can-i-calculate-if-a-satellite-is-currently-in-sunlight-or-eclipse-using-pye/40217#40217

https://space.stackexchange.com/questions/4339/calculating-which-satellite-passes-are-visible

It is commonly used in mission planning and adding this method would allow, in conjunction with the `find_events` function, calculating which events are potentially visible.

brandon-rhodes:
I agree that Skyfield would be improved by that feature! Let's leave this issue open until I or someone else can come up with some code.

brandon-rhodes:
(But as I won't be working on it immediately I guess I should un-assign myself so that other contributors will know that it's available to work on.)

jacoffey3:
I've found a solution for it. I'll see if I can at clean it up and get it PR worthy.

PR #369, opened 2020-04-30
Added is_sunlit method and docs for usage - Issue #362

jacoffey3:
This is a first pass. I tried to match as much as possible, but I'd be happy to restructure/reconfigure if you'd like. This is to add the feature I requested in https://github.com/skyfielders/python-skyfield/issues/362

I obviously pulled heavily from https://space.stackexchange.com/questions/37713/how-can-i-calculate-if-a-satellite-is-currently-in-sunlight-or-eclipse-using-pye/40217#40217 (as in copied and cleaned it up). 

I added documentation for usage, let me know if there is anything else I should add beyond that. (Wasn't sure about testing, since it's really just geometry and vector math)

### DIFF
--- a/skyfield/documentation/earth-satellites.rst
+++ b/skyfield/documentation/earth-satellites.rst
@@ -410,7 +410,7 @@ and will require providing ephemeris data.
 
 .. testsetup::
 
-  from datetime import datetime, timezone
+  from datetime import datetime, timezone, timedelta
   from skyfield.api import load, EarthSatellite
 
   # Set up the ISS as an example and load ephemeris.
--- a/skyfield/documentation/earth-satellites.rst
+++ b/skyfield/documentation/earth-satellites.rst
@@ -413,14 +413,14 @@ times you desire (when the TLE is still accurate):
 
   from skyfield.api import load
   satellite = by_name['ISS (ZARYA)']
-  de421.bsp = load('de421.bsp')
+  de421 = load('de421.bsp')
 
   # Define the times you are interested in
   minutes = range(0, 90, 10)
   times = ts.utc(2020, 4, 29, 0, minutes)
 
   # Calculate the sunlit vector
-  sunlit = satellite.is_sunlit(ephemeris=de421.bsp, times=times)
+  sunlit = satellite.is_sunlit(ephemeris=de421, times=times)
   for idx, time in enumerate(times):
       print(time.utc_jpl(), "{} is sunlit: {}".format(satellite.name,sunlit[idx]))
 
@@ -536,7 +536,7 @@ so it supports all of the standard Skyfield date methods:
     from skyfield.api import EarthSatellite
 
     text = """
-    GOCE                    
+    GOCE
     1 34602U 09013A   13314.96046236  .14220718  20669-5  50412-4 0   930
     2 34602 096.5717 344.5256 0009826 296.2811 064.0942 16.58673376272979
     """
--- a/skyfield/documentation/earth-satellites.rst
+++ b/skyfield/documentation/earth-satellites.rst
@@ -389,6 +389,68 @@ or else in dynamical coordinates of the date you specify.
 
 See :doc:`positions` to learn more about these possibilities.
 
+Finding when a satellite will be illuminated
+--------------------------------------------
+
+It is sometimes important to understand if the sun is
+currently illuminating the satellite.
+For instance, if it's after sundown in your location,
+and the satellite is still illuminated by the sun,
+you might be able to see that sunlight reflected with the naked eye.
+Additionally, this can be helpful when understanding
+satellite power and thermal cycles as it goes in and
+out of eclipse.
+
+Skyfield provides a simple geometric estimation for this
+with the :meth:`~skyfield.sgp4lib.EarthSatellite.is_sunlit()` method,
+which provides a boolean response for each corresponding time provided.
+This requires knowledge of the earth and sun,
+and will require providing ephemeris data.
+
+
+.. testsetup::
+
+  # Set up the ISS as an example and load ephemeris.
+  TLE = """ISS (ZARYA)
+  1 25544U 98067A   20120.54828704  .00001508  00000-0  35122-4 0  9994
+  2 25544  51.6452 229.1987 0001494 205.5139 329.7843 15.49330212224471"""
+
+  name, tle_1, tle_2 = TLE.splitlines()
+  ts = load.timescale()
+  satellite = EarthSatellite(tle_1, tle_2, name=name, ts=ts)
+  ephemeris = load('de421.bsp')
+
+Once you have the above defined, you can check
+if the satellite will be illuminated at whatever
+times you desire (when the TLE is still accurate):
+
+.. testcode::
+
+  # Define the times you are interested in
+  time = datetime(2020, 4, 29, tzinfo=timezone.utc)
+  times = ts.utc([time + timedelta(minutes=step) for step in range(0, 90, 10)])
+
+  # Calculate the sunlit vector
+  sunlit = satellite.is_sunlit(ephemeris=ephemeris, times=times)
+  for idx, time in enumerate(times):
+      print(time.utc_jpl(), "{} is sunlit: {}".format(satellite.name,sunlit[idx]))
+
+This produces a `sunlit` vector of booleans you can reference alongside your times:
+
+.. testoutput::
+
+  A.D. 2020-Apr-29 00:00:00.0000 UT ISS (ZARYA) is sunlit: False
+  A.D. 2020-Apr-29 00:10:00.0000 UT ISS (ZARYA) is sunlit: False
+  A.D. 2020-Apr-29 00:20:00.0000 UT ISS (ZARYA) is sunlit: True
+  A.D. 2020-Apr-29 00:30:00.0000 UT ISS (ZARYA) is sunlit: True
+  A.D. 2020-Apr-29 00:40:00.0000 UT ISS (ZARYA) is sunlit: True
+  A.D. 2020-Apr-29 00:50:00.0000 UT ISS (ZARYA) is sunlit: True
+  A.D. 2020-Apr-29 01:00:00.0000 UT ISS (ZARYA) is sunlit: True
+  A.D. 2020-Apr-29 01:10:00.0000 UT ISS (ZARYA) is sunlit: True
+  A.D. 2020-Apr-29 01:20:00.0000 UT ISS (ZARYA) is sunlit: False
+
+You can see when it is sunlit at the given times!
+
 Avoid calling the observe method
 --------------------------------
 
@@ -485,7 +547,7 @@ so it supports all of the standard Skyfield date methods:
     from skyfield.api import EarthSatellite
 
     text = """
-    GOCE                    
+    GOCE
     1 34602U 09013A   13314.96046236  .14220718  20669-5  50412-4 0   930
     2 34602 096.5717 344.5256 0009826 296.2811 064.0942 16.58673376272979
     """
--- a/skyfield/documentation/earth-satellites.rst
+++ b/skyfield/documentation/earth-satellites.rst
@@ -406,35 +406,21 @@ with the :meth:`~skyfield.sgp4lib.EarthSatellite.is_sunlit()` method,
 which provides a boolean response for each corresponding time provided.
 This requires knowledge of the earth and sun,
 and will require providing ephemeris data.
-
-
-.. testsetup::
-
-  from datetime import datetime, timezone, timedelta
-  from skyfield.api import load, EarthSatellite
-
-  # Set up the ISS as an example and load ephemeris.
-  TLE = """ISS (ZARYA)
-  1 25544U 98067A   20120.54828704  .00001508  00000-0  35122-4 0  9994
-  2 25544  51.6452 229.1987 0001494 205.5139 329.7843 15.49330212224471"""
-
-  name, tle_1, tle_2 = TLE.splitlines()
-  ts = load.timescale()
-  satellite = EarthSatellite(tle_1, tle_2, name=name, ts=ts)
-  ephemeris = load('de421.bsp')
-
-Once you have the above defined, you can check
-if the satellite will be illuminated at whatever
+You can check if the satellite will be illuminated at whatever
 times you desire (when the TLE is still accurate):
 
 .. testcode::
 
+  from skyfield.api import load
+  satellite = by_name['ISS (ZARYA)']
+  de421.bsp = load('de421.bsp')
+
   # Define the times you are interested in
-  start_time = datetime(2020, 4, 29, tzinfo=timezone.utc)
-  times = ts.utc([start_time + timedelta(minutes=step) for step in range(0, 90, 10)])
+  minutes = range(0, 90, 10)
+  times = ts.utc(2020, 4, 29, 0, minutes)
 
   # Calculate the sunlit vector
-  sunlit = satellite.is_sunlit(ephemeris=ephemeris, times=times)
+  sunlit = satellite.is_sunlit(ephemeris=de421.bsp, times=times)
   for idx, time in enumerate(times):
       print(time.utc_jpl(), "{} is sunlit: {}".format(satellite.name,sunlit[idx]))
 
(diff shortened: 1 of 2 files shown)


skyfield/documentation/earth-satellites.rst
```diff
@@ -389,6 +389,68 @@ or else in dynamical coordinates of the date you specify.
...
+  ephemeris = load('de421.bsp')
+
+Once you have the above defined, you can check
+if the satellite will be illuminated at whatever
+times you desire (when the TLE is still accurate):
+
+.. testcode::
+
+  # Define the times you are interested in
+  time = datetime(2020, 4, 29, tzinfo=timezone.utc)
```
jacoffey3, line 430:
```suggestion
  start_time = datetime(2020, 4, 29, tzinfo=timezone.utc)
```

jacoffey3:
Looking into the failing test

skyfield/documentation/earth-satellites.rst
```diff
@@ -389,6 +389,68 @@ or else in dynamical coordinates of the date you specify.
...
+Skyfield provides a simple geometric estimation for this
+with the :meth:`~skyfield.sgp4lib.EarthSatellite.is_sunlit()` method,
+which provides a boolean response for each corresponding time provided.
+This requires knowledge of the earth and sun,
+and will require providing ephemeris data.
+
+
+.. testsetup::
+
+  # Set up the ISS as an example and load ephemeris.
```
jacoffey3, line 413:
```suggestion
      from datetime import datetime, timezone
      from skyfield.api import load, EarthSatellite
  
      # Set up the ISS as an example and load ephemeris.
```

brandon-rhodes:
So having just re-read `compute_limb_angle()`, I'm less convinced than I was that it properly handles Earth's ellipsoidal-ness. It looks like it considers the Earth a sphere, but that it doesn't need high resolution because it's only ever used as a lower-bound cutoff for whether to include the Earth when computing how the gravity of various objects deflect the incoming light from a distant object.

So let me go look at SPICE again for an equivalent.

jacoffey3:
Thanks for the feedback! 

The whitespace was my auto-linter. Sorry! I can backtrack them. 

I'll checkout using the `compute_limb_angle()` to do the calc. It's not my math, so I am not really looking to dive into redoing it with this PR (if the skyfield method doesn't work). Totally understand if you need it done though. 

I think distinguishing umbral vs penumbral would be a great add-on, but I don't know if I'm willing to dive in at the moment. I'll look into and see how difficult it would be. 

I understand the desire the put it in the almanac package, but I'm unsure if it will satisfy the same use cases. Wouldn't I need to make an assumption on the step size  to do that calculation anyways? I'd assume it would be something like: 

```python
times = almanac.sunlit_times(t0=start_time, t1=end_time, sat=EarthSatellite())
enum = ["penumbral_start":0, "umbral_start":1, "umbral_end":2, "penumbral_end":3]
print(times)

([time1, enum],[time2, enum]...)
```

But what would that time need to be accurate to? 
Doing it this way would allow adding other object more easily down the road, but I'm really only interested in the satellite case. It would also make it look more like your `EarthSatellite.find_events() method...`

skyfield/sgp4lib.py
```diff
@@ -27,6 +27,7 @@
 _infs = array(('-inf', 'inf'), float)
 _ts = Timescale(array((_infs, (0.0, 0.0))), _infs, array((37.0, 37.0)))
 
+
```
jacoffey3, line 30:
```suggestion
```

skyfield/documentation/earth-satellites.rst
```diff
@@ -485,7 +536,7 @@ so it supports all of the standard Skyfield date methods:
     from skyfield.api import EarthSatellite
 
     text = """
-    GOCE                    
+    GOCE
```
jacoffey3, line 539:
```suggestion
    GOCE                    
```

skyfield/sgp4lib.py
```diff
@@ -243,8 +268,10 @@ def below_horizon_at(t):
         i = jd.argsort()
         return ts.tt_jd(jd[i]), v[i]
 
+
```
jacoffey3, line 271:
```suggestion
```

skyfield/sgp4lib.py
```diff
@@ -243,8 +268,10 @@ def below_horizon_at(t):
...
 _second = 1.0 / (24.0 * 60.0 * 60.0)
 
+
```
jacoffey3, line 274:
```suggestion
```

skyfield/sgp4lib.py
```diff
@@ -262,6 +289,7 @@ def theta_GMST1982(jd_ut1):
     theta_dot = (1.0 + dg * _second / 36525.0) * tau
     return theta, theta_dot
 
+
```
jacoffey3, line 292:
```suggestion
```

skyfield/documentation/earth-satellites.rst
```diff
@@ -536,7 +536,7 @@ so it supports all of the standard Skyfield date methods:
...
```
jacoffey3, line 539:
```suggestion
    GOCE                    
```

jacoffey3:
On another note, I finally wrapped my head around your doc test system and got it passing. I'm stoked.

brandon-rhodes:
> On another note, I finally wrapped my head around your doc test system and got it passing. I'm stoked.

Nice! I'm glad you got it working.

On my end, I have finally found the SPICE code for figuring out if a vector (like the satellite-to-Sun vector) intercepts an ellipsoid (like the Earth):

https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/surfpt.html

By downloading any copy of SPICE you should be able to try reading the routine's source code. (Or maybe someone has SPICE up on GitHub?) Folks say the FORTRAN is easiest to read, since the C is translated from the FORTRAN by a program that doesn't try to make the C particularly beautiful.

Take a look and let me know if the vector computations look doable for you. If so, go for it! If not, then I'll do a code review of your current diff, and maybe we can land it with an underscore in front of its name for temporary use until we have the higher-accuracy version working.

The difference between the Earth's polar and central radii is small, about 21 meters:

```
from skyfield import api
ts = api.load.timescale(builtin=True)

x, y, z = api.Topos('0 N', '0 W').itrf_xyz().km
radius_equator_m = x

x, y, z = api.Topos('90 N', '0 W').itrf_xyz().km
radius_pole_m = z

print(radius_equator_m - radius_pole_m)
```

An Earth satellite at right angles to a difference of 21 meters would not even notice; the difference would be about 0.003 seconds earlier or later. My worry is, what if there are slower higher-orbit satellites that come into or out of eclipse by hitting the shadow at a steep glancing angle? In that case 21 meters could possibly be more significant.

And I don't know which is more work: starting to draw hypothetical spacecraft orbits to convince myself that the 21 meters never matters, or simply implementing the math for `surfpt`. The math looks pretty simple so I'm leaning that way. :)

Let me know what you decide!

jacoffey3:
> And I don't know which is more work: starting to draw hypothetical spacecraft orbits to convince myself that the 21 meters never matters, or simply implementing the math for surfpt. The math looks pretty simple so I'm leaning that way. :)

IMO, it's not going to matter for most satellites for either use case: 

1. For power/thermal profiling, your unknowns around attitude, thermal behavior, solar panel efficiency, etc all have some degree of unknowns associated with them, and it would fall within that margin. 

2. For looking for visible satellites, you're obviously kinda shooting in the dark (literally), because the reflectivity and attitude of the spacecraft and entirely unknown. It's more for "huh I wonder if I can see this?", or "I'm getting this point source in my data I didn't expect to, can I potentially use this to filter it out?". Where this difference would really fall well within the margin of error. 

As a result, I would suggest that I change it to put it in the `almanac.py` package like I gave the example for, but only support the `EarthSatellites` use case for now, document that we are assuming a spherical earth for this calculation, and put a TODO on the list to extend it to support ellipsoid models and other objects (like Jupiter, as you stated before). What do you think?

brandon-rhodes:
I agree that it's fine to land it with the spheroid model — but if it's satellite-specific then it probably belongs as a method on satellites rather than in Almanac, which is supposed to be routines that work on any object.

Let me go take a look at the code, then, and see if I have any comments!

jacoffey3:
I also have very limited C experience and no FORTRAN, so I'm not sure if I want to learn it _right now..._

skyfield/sgp4lib.py
```diff
@@ -153,6 +153,30 @@ def ITRF_position_velocity_error(self, t):
...
         return rITRF, vITRF, error
 
+    def is_sunlit(self, ephemeris, times):
+        """Calculates if the satellite is currently sunlit (True/False) at each time in times. Returns an array of booleans."""
+        Re = 6378.137  # Equatorial radius of the Earth in km
+        earth = ephemeris['earth']
+        sun = ephemeris['sun']
+        sat = earth + self
+
+        sunpos, earthpos, satpos = [
```
brandon-rhodes, line 163: Let's pivot these from a list to a plain set of 3 assignments. That way the correspondences are clearer, the code will run faster, and it will take as many lines of code anyway.

jacoffey3:
Sounds good! I enjoy the discussion and hope it proves useful :)

skyfield/sgp4lib.py
```diff
@@ -153,6 +153,30 @@ def ITRF_position_velocity_error(self, t):
...
+            thing.at(times).position.km for thing in (sun, earth, sat)
+        ]
+        sunearth, sunsat = earthpos-sunpos, satpos-sunpos
+        sunearthnorm, sunsatnorm = [
```
brandon-rhodes, line 167: Let's also unwrap this loop, for simplicity in both reading and when Python executes it. And I think you can use `skyfield.fuctions.length_of()` here instead of the raw formula — as well as on the following lines. It will make the code easier for me, at least, to read and maintain. Thanks!

skyfield/sgp4lib.py
```diff
@@ -153,6 +153,30 @@ def ITRF_position_velocity_error(self, t):
...
+            vec/sqrt((vec**2).sum(axis=0)) for vec in (sunearth, sunsat)
+        ]
+        angle = arccos((sunearthnorm * sunsatnorm).sum(axis=0))
+        sunearthdistance = sqrt((sunearth**2).sum(axis=0))
+        sunsatdistance = sqrt((sunsat**2).sum(axis=0))
+        limbangle = arctan2(Re, sunearthdistance)
+        sunlit = []
+        for idx, value in enumerate(angle):
```
brandon-rhodes, line 175: Happily, you should be able to omit this loop by computing the trues and falses directly with NumPy. Check out their docs on boolean arrays.

skyfield/sgp4lib.py
```diff
@@ -153,6 +153,30 @@ def ITRF_position_velocity_error(self, t):
         rITRF, vITRF = TEME_to_ITRF(t.ut1, rTEME, vTEME)
         return rITRF, vITRF, error
 
+    def is_sunlit(self, ephemeris, times):
+        """Calculates if the satellite is currently sunlit (True/False) at each time in times. Returns an array of booleans."""
+        Re = 6378.137  # Equatorial radius of the Earth in km
```
brandon-rhodes, line 158: Instead of hard-coding this value, go ahead and use `ERAD` from the `skyfield.constants` module.

skyfield/sgp4lib.py
```diff
@@ -153,6 +153,30 @@ def ITRF_position_velocity_error(self, t):
...
+            vec/sqrt((vec**2).sum(axis=0)) for vec in (sunearth, sunsat)
+        ]
+        angle = arccos((sunearthnorm * sunsatnorm).sum(axis=0))
+        sunearthdistance = sqrt((sunearth**2).sum(axis=0))
+        sunsatdistance = sqrt((sunsat**2).sum(axis=0))
+        limbangle = arctan2(Re, sunearthdistance)
+        sunlit = []
+        for idx, value in enumerate(angle):
+            sunlit.append(((angle[idx] > limbangle[idx]) or (
+                sunsatdistance[idx] < sunearthdistance[idx])))
```
brandon-rhodes, line 177: Add a comment noting that we are omitting the case that a satellite below the Earth's surface is not sunlit, since that's the third general case that these angles can add up to, but we're conspicuously giving ourselves permission to ignore it. :)

skyfield/sgp4lib.py
```diff
@@ -153,6 +153,30 @@ def ITRF_position_velocity_error(self, t):
...
+            thing.at(times).position.km for thing in (sun, earth, sat)
+        ]
+        sunearth, sunsat = earthpos-sunpos, satpos-sunpos
+        sunearthnorm, sunsatnorm = [
+            vec/sqrt((vec**2).sum(axis=0)) for vec in (sunearth, sunsat)
+        ]
+        angle = arccos((sunearthnorm * sunsatnorm).sum(axis=0))
+        sunearthdistance = sqrt((sunearth**2).sum(axis=0))
+        sunsatdistance = sqrt((sunsat**2).sum(axis=0))
+        limbangle = arctan2(Re, sunearthdistance)
```
brandon-rhodes, line 173: The diagram I've sketched suggests that `arctan2()` does not work here? I'll try it in ASCII art (gulp!):

```
Sun

|\ <--- angle
| \
|  \
|   \   sunearthdistance
|_   \
|_|___\
       Earth center

 Earth
 radius
```
Feel free to follow up with another diagram if I've imagined the triangle a bit wrong.

brandon-rhodes:
And just minutes after sharing my ASCII art diagram here, I ran across:

image

Just in case we needed any more diagram inspiration :) [From [here](https://github.com/newtonee/Visibility-of-the-Sun/blob/master/When_the_Sun_is_visible_.ipynb)]

jacoffey3:
@brandon-rhodes I'll check out the comments! Got tied up so it might be a few days before I'm able to revisit.

brandon-rhodes:
> @brandon-rhodes I'll check out the comments! Got tied up so it might be a few days before I'm able to revisit.

Thanks for letting me know! I'm not blocked on this particular PR, so no problem — I'll just watch for when you get time and it shows activity again.

brandon-rhodes:
@jacoffey3 — Over in another issue discussing the Earth occulting things, I have tried implementing the formula for finding the nearest intersection of a line with a sphere; it’s in this comment here:

https://github.com/skyfielders/python-skyfield/issues/377#issuecomment-634168823

Would it be possible for the interior math here in your routine to use the same formula to ask “does the satellite → Sun vector intersect the surface of the Earth?” Since the routine is pure vector math, it avoids the expense of any trigonometry, and might even handle the case of a satellite below the Earth's surface (but I'm not sure, maybe it doesn't).

brandon-rhodes:
@jacoffey3 — I am thinking of making a release today, so I think I'll go ahead and land this as-is, and then adjust the interior code to use the new routine that avoids any questions of trigonometry. Thanks for suggesting this addition and writing documentation that tests it! It will assure that the new code gets the same answers as the old. :)

jacoffey3:
@brandon-rhodes Sorry I dropped off! I'm back around, and I agree using that formula is best. My apologies for orphaning the PR. I greatly appreciate you going ahead and landing it.

brandon-rhodes:
@jacoffey3 — There is one adjustment I might make to the API. If someone takes a satellite, computes some positions for it, and then wants to know when it's in sunlight, then the current routine seems to present a problem: it can't accept those positions as input, but instead might need to compute them all over again a second time. Do you agree that the method’s design has that snag, or is there a way you thought of around it?

I'm wondering if this should be a method on `Geometric` positions instead. What would you think of that? Then you could take a satellite, `.at(t)` it to learn its position, and then `.is_sunlit()` the position without its x,y,z having to be re-computed.

jacoffey3:
@brandon-rhodes It would definitely be the more efficient way. Especially since if you didn't want to deal with storing it locally you can do the exact same process as the current method, just a little bit more verbose: `satellite.is_sunlit(satellite.at(t))` should get the same result. The only other thing is if I have an array of times, can I do that and still get the array of booleans back? 

Also since that would make it so that it's just a thin layer that makes the sun/earth assumption for you, does it make sense to have it as a separate api call? Or should it become an example about how to use your new underlying method for the use case "determining if a satellite is sunlit"?

brandon-rhodes:
@jacoffey3 — I am puzzled that our routines do not agree for the example you set up in the documentation. Having looked at the numbers from several directions, I am increasingly convinced that your 2020-04-29 ISS example is tackling a situation in which (at least as predicted by the TLE) the ISS is underground, and orbiting the Earth about 150 kilometers below its surface.

In such a situation I am satisfied that my routine says "shadow" for the whole time period covered by your example, since to be underground is to be in eclipse.

Your algorithm, on the other hand, I'm guessing treats the Earth as a flat disc, and so the ISS would be passing in front of and behind the flat disc regardless of how tiny its orbit becomes around the Earth’s core.

If you could glance at the numbers and confirm my suspicion, it would give me more confidence in replacing the trig algorithm with the vector algorithm. (And it's fine if you don't have time; it would just be fun to have a second opinion in case you do have time!)

jacoffey3:
@brandon-rhodes that seems very strange indeed. As for the TLE I just pulled it from Celestrak to work the example. I will try to revisit as soon as possible.

brandon-rhodes:
No rush! But I will be interested whether you also get a very low altitude for the ISS from those elements. I wonder if it's possible that amidst my cutting and pasting I used much older elements for the satellite than your test was supposed to be run against?

In any case, I'm very happy with the feature an am glad you offered it!

jacoffey3:
@brandon-rhodes So first of all, yes you're correct, my initial was treating the earth as a flat disk, naively assuming "if the sat is closer than the earth center, it's in sun". 

But for the TLE, the test is actually pulling directly from what was already there at the beginning of the file. I think it's below the earth due to 6 years of propagation of a very _very_ stale TLE. I'd suggest moving the example to be at the same time as the rest of the examples, then it should make more sense :)

brandon-rhodes:
@jacoffey3 — That makes perfect sense, thanks for double-checking me!

And thanks again for pushing for this feature, even ahead of correct treatment of the Earth as an oblate spheroid; I think you are correct that for the typical use case, the different will be negligible, and thanks to your argument users have this feature now rather than later.

Issue #394, opened 2020-06-22
general perturbations (GP) class to replace TLE format


craigim:
Both celestrak.com and space-track.org have recently (past week) put up banners that the TLE is becoming obsolete because they're running out of 5-digit NORAD IDs. Both are promoting the [GP class](https://www.celestrak.com/NORAD/documentation/gp-data-formats.php) as the new preferred format for getting satellite SPG4 inputs.

This is a feature request that SkyField support inputs for satellites in the new format, both as XML and as JSON.

glangford:
I am working on some early test code to ingest the XML and TLE, run the two through Skyfield and compare the results. Seems to be mostly equivalent so far (using just ISS as an example). I can share the test code if it's useful. 

It is possible to create EarthSatellite objects from OMM XML without changes to Skyfield itself if that is needed in the short term.

mworion:
Would be great to see it soon in Skyfield as load capability. As SGP4 already has the capability to transform a SATREC object back to TLE I would like to have the GP format to be supported. My mount could be programmed with TLE elements so a conversion is needed.

Michel

ovillami:
> 
> 
> I am working on some early test code to ingest the XML and TLE, run the two through Skyfield and compare the results. Seems to be mostly equivalent so far (using just ISS as an example). I can share the test code if it's useful.
> 
> It is possible to create EarthSatellite objects from OMM XML without changes to Skyfield itself if that is needed in the short term.

Hi Glangford,

I would be interested in your test code, if you could share it, it would be great.

glangford:
Hi @ovillami , there is a pull request with code and test data to compare TLE and XML. Suggestions welcome.

brandon-rhodes:
Thanks for the discussion, everyone! I am returned from a hiking trip and will be happy to see progress on this. I’ll continue the conversation over on the pull request for now.

brandon-rhodes:
The pull request was resolved back on January 8, and I forgot to return and close this issue that inspired it! Better late than never, I suppose.

PR #398, opened 2020-06-28
Test TLE vs XML for similar Position (Issue #394)

glangford:
Test/Evaluation code  - for Stations in stations.txt and the XML equivalent data published by Celestrak, create 'twin' `EarthSatellite` objects using each source and verify that they have a similar position at a set time. For the included TLE and XML files, the positions of all Stations are essentially identical.


brandon-rhodes:
@glangford — Thank you for drafting this Python code! I suspect that it will not only be important to Skyfield users, but to all users of the underlying `sgp4` library for Python that Skyfield uses. Can we move this pull request over to its repository instead?

https://github.com/brandon-rhodes/python-sgp4/

There, it will be able to benefit other projects and not just Skyfield. Thanks!

glangford:
> I suspect that it will not only be important to Skyfield users, but to all users of the underlying sgp4 library for Python that Skyfield uses. Can we move this pull request over to its repository instead?

@brandon-rhodes Yes, I can do that. Please let me know of any suggestions (are you ok with the test files as they are?)

I will keep this pull request open for a while in case there are additional comments.

glangford:
Closing since the code has been moved to the underlying sgp4 library.

brandon-rhodes:
Thanks!

Issue #389, opened 2020-06-10
Skyfield far slower than ephem


Sec42:
I wrote a small script to match sightings with satellites via their TLE.

The script using skyfield is about 80x slower than the same using ephem.

My question is: can this be improved? Am I using skyfield wrong here?

```
#!/usr/bin/python3
import sys
import datetime
from skyfield.api import load, utc, Topos

satlist = load.tle_file(sys.argv[1])
ts = load.timescale(builtin=True)


tim=float(sys.argv[2])/1000.
lat=float(sys.argv[3])
lon=float(sys.argv[4])
alt=int(sys.argv[5])*1000

time_t = datetime.datetime.utcfromtimestamp(tim)
time_t = time_t.replace(tzinfo=utc)
t = ts.utc(time_t)
mysat=Topos(latitude_degrees=lat, longitude_degrees=lon, elevation_m=alt)
my=mysat.at(t)

minsep=9e9
sname=None
for sat in satlist:
    pos=sat.at(t)

    sep=pos.separation_from(my).degrees
    if sep<minsep:
        sname=sat.name
        minsep=sep

print("Minimum: %4.2f: %s"%(minsep/1000,sname))
```
 
vs. similar code with ephem:

```
#!/usr/bin/python3
import sys
import math
import datetime
import ephem

degrees_per_radian = 180.0 / math.pi

satlist = loadTLE(sys.argv[1])

tim=float(sys.argv[2])/1000.
lat=float(sys.argv[3])
lon=float(sys.argv[4])

t = datetime.datetime.utcfromtimestamp(tim)

minsep=9e9
sname=None
for sat in satlist:
    sat.compute(t)

    sep=ephem.separation((sat.sublong,sat.sublat),(lon/degrees_per_radian,lat/degrees_per_radian))*degrees_per_radian
    if sep<minsep:
        sname=sat.name
        minsep=sep

print("Minimum: %4.2f: %s"%(minsep,sname))
```

As a datapoint: using a TLE file with 75 satellites, and running in a 1000x loop, I get runtimes of 0.17s for ephem vs. 14s for skyfield.

Is a way to optimize/speed up what I'm doing with skyfield?

Thanks,
   Sec

brandon-rhodes:
Yes, alas, saving 0.8 seconds — which is normally noticeable — is in this case not a substantial improvement!

So, let me at least quickly answer your original question:

[...]

Issue #367, opened 2020-04-27
Vectorized find_events computation


jvarada:
Is there a way to call the find_events method in some sort of vectorized fashion to compute the visibility windows of a satellite at multiple positions (I'm imagining something like passing a list of  Topos objects for a given start/end time and altitude)?

brandon-rhodes:
There are not built-in routines for either kind of vectorization, I don't think, and I'm not sure how large any speed increase would be — the underlying satellite position computation doesn't use vector math, but (hopefully on modern machines at least) is a C++ routine that is called *n* times if *n* positions are needed.

Your second case — same location, many satellites — is interesting, though, because the underlying C++ does at least have the ability to produce NumPy arrays from a loop over *n* satellite objects rather than having to create dozens or hundreds of separate Python `float` objects. Let me go see what I can put together!

brandon-rhodes:
So: again, your first situation is not likely to support a vectorized speedup, because the number of computations necessary to do the computation is large compared to the number of satellites, so even if there were a way to somehow run it as vectors (and there's probably not? feel free to suggest a possible vectorization if you can think of one and we can consider it!), you'd only eliminate one iteration per satellite which I think is many fewer than the number of positions each satellite needs to be computed for to get rise and set times.

But for asking at time `t` which satellites are above the horizon can benefit directly from the underlying C++ library's ability to consider a vector of satellites. And better yet, if we want to be low-level we can skip converting all the vectors into Skyfield's base ICRF reference frame and can instead leave all the vectors in an Earth-fixed reference frame. See if, given a date and TLE file, this script produces the altitude in degrees (ignoring atmospheric refraction) of all the satellites — and, if you have a few minutes, try comparing its runtime to what you've been doing already:

```
# Generate Earth location.

import numpy as np
from skyfield import api
from skyfield.functions import length_of, dots
from skyfield.sgp4lib import TEME_to_ITRF
from sgp4.api import SatrecArray

ts = api.load.timescale(builtin=True)
t = ts.utc(2020, 4, 28, 11, 44)
topos = api.Topos('40.8939 N', '83.8917 W')
p = topos.itrf_xyz().km

# Build satellite array.

sats = api.load.tle_file('stations.txt')
print(sats[0])

sat_array = SatrecArray([sat.model for sat in sats])

jd = np.array([t._utc_float()])
fr = jd * 0.0
e, r, v = sat_array.sgp4(jd, fr)

[...]

Issue #188, opened 2018-07-06
Calculation speed of skyfield orbit propagation for multiple EarthSatellite objects loaded from TLE


dsholes:
Apologies if this is too similar to #30 and #183. I decided to investigate the bottleneck for orbit propagation (position and velocity) for multiple `skyfield.sgp4lib.EarthSatellite` objects, using `EarthSatellite.at()`

# Summary:
- Bottleneck is related to `skyfield.positionlib.ITRF_to_GCRS2()`, which relies on "expensive" functions from `skyfield.nutationlib` and `skyfield.precessionlib` to calculate important attributes for `skyfield.timelib.Time` object.
    - If future work related to Numba-jit takes place ([mentioned in Issue #30](https://github.com/skyfielders/python-skyfield/issues/30#issuecomment-58933757)), it would be nice to address `skyfield.nutationlib` and `skyfield.precessionlib`

- In the meantime, for those who are interested in potential speed boosts, I have two ideas for reducing computation time for multiple satellite orbits
    - [Option 1](#option-1) - Ideally, we would use the same `skyfield.timelib.Time` object for every `skyfield.sgp4lib.EarthSatellite` object that we want to propagate. Then we could pre-compute `skyfield.timelib.Time.MT` and `skyfield.timelib.Time.gast` (the "expensive" attributes mentioned above), and the results would be cached for all of the proceeding satellite propagations.
        - Since accuracy of the sgp4 propagation rapidly degrades before/after TLE epoch, users must understand the risks associated with using Times "far away" from TLE epoch for each satellite, since TLE epoch is different for each satellite in the Celestrak NORAD elements (e.g. 'http://celestrak.com/NORAD/elements/planet.txt')
        - For 15 satellites and 5760 orbit positions, using a single time array (`skyfield.timelib.Time` object) for all satellites and pre-computing/caching `t.MT` and `t.gast` **reduces computation time from 35 sec to 5 sec**.
    - [Option 2](#option-2) - Another idea is to allow users (with a keyword argument) to return position and velocity in TEME or ITRF reference frames instead of GCRS, for reduced computation time. This allows for even greater speed (**~3 sec** for 15 satellites and 5760 orbit positions), and doesn't restrict user to single time array (`skyfield.timelib.Time` object) for all satellites. The obvious drawback is some users need pos and vel in GCRS

```python
from skyfield.api import Topos, load
import numpy as np
import time
```

```python
# Orbit propagation parameters, for 5760 orbit positions over time
[...]

PR #401, opened 2020-07-07
Prototype support for parallel SGP4 across multiple satellites

glangford:
This is an initial prototype to test package the parallel SGP4 capability for computing "n satellites x m times" into one routine in Skyfield. 

The new function `positions_for(satellites, times)` takes an array of `EarthSatellite` objects and an array of `Time`, calls SGP4 to calculate states in parallel, and converts the TEME coordinates to geocentric positions. Coordinate conversion follows the methods used today.

The brief test code shows that this version gives the same `Position` results as calling `EarthSatellite.at()`.

This is a separate function for early development but perhaps it could evolve into a `@classmethod`. Suggestions welcome on what the API should look like.

Questions

- In combination with other optimizations that have been discussed around coordinates, should we pursue this further (eg. for #188, #367, #389) or is it already in development ?

- Does `.sgp4()` release the GIL? As @dsholes suggested,
> ...Then we could pre-compute skyfield.timelib.Time.MT and skyfield.timelib.Time.gast

in this case, pre-computation might be done while n x m SGP4 was running.

### DIFF
--- a/design/sgp4_parallel.py
+++ b/design/sgp4_parallel.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+"""
+The SGP4 library has the ability to process an array of satellites across an
+array of times. What could support for this look like in Skyfield, combining
+optimized parallel calculation (in the C++ version of SGP4) with coordinate
+conversions from TEME?
+"""
+import numpy as np
+from sgp4.api import SatrecArray, SGP4_ERRORS, accelerated
+from skyfield.api import load, utc, EarthSatellite
+from skyfield.constants import AU_KM, DAY_S, AU_M
+from skyfield.sgp4lib import TEME_to_ITRF
+from skyfield.positionlib import ITRF_to_GCRS2, build_position
+from skyfield.vectorlib import ObserverData
+
+# Run SGP4 in parallel across n satellites x m times
+def positions_for(satellites, times):
+    sat_array = SatrecArray( [s.model for s in satellites] )
+    jd = times._utc_float()
+    e, r, v = sat_array.sgp4(jd, np.zeros_like(jd))
+    assert len(satellites) == np.shape(r)[0]
+    for index in range(r.shape[0]):
+        errors = e[index]
+        messages = [SGP4_ERRORS[error] if error else None for error in errors ]
+        # Adapted from _position_and_velocity_TEME_km(), ITRF_position_velocity_error()
+        rTEME = np.divide(r[index].T, AU_KM)
+        vTEME = np.divide(v[index].T, AU_KM / DAY_S)
+        rITRF, vITRF = TEME_to_ITRF(times.ut1, rTEME, vTEME)
+        rGCRS, vGCRS = ITRF_to_GCRS2(times, rITRF, vITRF)
+        # Adapted from VectorFunction at()
+        observer_data = ObserverData()
+        observer_data.gcrs_position = rGCRS
+        target = -100000 - satellites[index].model.satnum
+        position = build_position(rGCRS, vGCRS, times, 399, target, observer_data)
+        position.message = messages
+        yield position
+
+# ------------------------------------------------------------------------- #
+
+ISS_TLE = """1 25544U 98067A   20187.34541214  .00002866  00000-0  59793-4 0  9991
+2 25544  51.6454 258.5877 0002990  76.8215 328.5773 15.49181247234835
+"""
+
+KESTREL_TLE = """1 42982U 98067NE  20187.19650118  .00008987  00000-0  76592-4 0  9999
+2 42982  51.6323 190.7341 0002835 141.7806 218.3393 15.70424148153865
+"""
+
+def main():
+    if not accelerated:
+        print("SGP4 library is not accelerated")
+        return
+    test_iss()
+    test_two_sats()
+    test_many_sats()
+
+def test_iss():
+    ts = load.timescale(builtin=True)
+    L1, L2 = ISS_TLE.splitlines()
+    iss = EarthSatellite(L1, L2)
+
+    trange = range(0, 60, 10)
+    times = ts.utc(2020, 7, 5, 12, 0, trange)
+    print("positions_for():")
+    satellites = [iss, iss]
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
I'm happy to have this stay open, if you want to re-open at any point, though it's true that I wasn't able to get to it quickly.

glangford:
@brandon-rhodes I closed it because I stepped through some of vectorlib in the debugger and realized the prototype was inadequate. I need to learn more of the underpinnings and make changes at some point!

brandon-rhodes:
Was the obstacle the need for a vector to hold positions for both *n* times but also *m* satellites, or was the problem different than that?

glangford:
Rather than pushing support for n x m down into `VectorFunction`, I was thinking of hoisting it out. The thought process was that VectorFunction as a lower level class ideally wouldn't have to sweat about multiple satellites.

Following the canonical method
> "Every time you call this vector sum’s at() method, it will first compute the satellite’s position, then your own position, and finish by subtracting them."

Instead, conceptually this higher level function could receive just one location, and an array of satellites - not an array of `VectorSum`. Then it would know to run parallel sgp4 *once* for the satellites, and then drive the conversion afterward. So this is less work for the caller (no need to repeat `difference = satellite - location` many times) and more efficient to compute. But I haven't puzzled through it all yet.  :) Does that make sense?

brandon-rhodes:
Thanks for the additional thoughts! I suppose I had vaguely thought that `VectorFunction` already provides support for as many extra dimensions as the programmer wants to use, and that nothing additional needed to be pushed down into it. But you probably have more experience with it in that situation than I do — how are you imagining `VectorFunction` being changed? I didn't see any adjustments to it here on this PR so a specific example might be helpful.

glangford:
I haven't got that far yet, I only did the basics in the original PR and assumed that `VectorFunction` would work out. But then it didn't seem right to have the caller supply an array of objects all subtracting the same location, and so I am down the rabbit hole still.

brandon-rhodes:
> But then it didn't seem right to have the caller supply an array of objects all subtracting the same location…

Could you sketch how that would look? Upon first impression it sounds quite efficient and NumPy-accelerated to have *m* satellite positions all then subtract the same observer position, so it’s how Skyfield should work.

glangford:
More specifically:

The basic interface in the original PR was

`def positions_for(satellites, times):`

and I was thinking along the lines of

`def altaz_for(location, satellites, times):`
 
The idea was to only calculate the negative location once (under the covers) and apply to all satellites, rather than do this:
```
def _at(self, t):
        p, v = 0.0, 0.0
        for segment in self.positives:
            p2, v2, gcrs_position, message = segment._at(t)
            p += p2
            v += v2
        for segment in self.negatives:
            p2, v2, gcrs_position, ignored_message = segment._at(t)
            p -= p2
            v -= v2
        return p, v, gcrs_position, message
```

brandon-rhodes:
I think your original interface might work fine. What if:
```
def positions_for(satellites, times):
```
— returned a geocentric position vector with NumPy dimensions `(3, t_count, sat_count)`. If we then took a normal observer position for that time, with dimensions `(3, t_count)`, and quickly recast it to dimensions `(3, t_count, 1)`, then I think NumPy would do the subtraction exactly as you want.

In which case, your original point now becomes clearer, the `_at()` method you just quoted would probably need to add the extra `(…, 1)` dimensions on-the-fly, because there's no way currently to tell an Earth position "remember, when your `_at()` is called, throw in one or two extra dimensions because I'm going to be combining it with multiple objects along some other axis."

Similar logic already exists to let `separation_from()` compare, say, *n* positions of Jupiter with a single point in the sky:

https://github.com/skyfielders/python-skyfield/blob/master/skyfield/positionlib.py#L251

Hopefully it could be implemented to not slow down `_at()` too much in the common case?

glangford:
That's a good thought, and thanks for sharing the helpful code sample. I have lots to learn about numpy still, so I will need to read more code and digest this further.   :)

glangford:
It looks like VectorSum will call `_at()` for all of its segments, which is not what we want to do in the  multi satellite case. What we need to have happen is to just do the subtraction of the `Topos` location for all the Satellite positions we already have in hand. So VectorSum as it exists today doesn't really fit.

I am thinking the easiest way to tackle this is in the one new routine for the moment
- compute the TEME position for all Satellites in parallel
- compute the position of the viewing location (passed in as a new parameter)
- loop through the TEME positions
   - convert to GCRS (same as before)
   - subtract the viewing location (in the way that `VectorSum _at()` would do for `self.negatives`)
   - yield the computed result (which I guess will be a `Geometric`)

Maybe this gets refactored later. Anyway, that's where I'm currently at.

brandon-rhodes:
I am imagining a final public API that's a Skyfield `SatelliteArray` that would support, parallel to the existing `(sat - observer).at(t)`, the maneuver `(sat_array - observer).at(t)`. But if a better first step given your needs is a high-level single-purpose routine that builds the satellite array and arranges the steps differently, it would still be a useful thing to have working, and I'm interested in seeing the result.

glangford:
> I am imagining a final public API that's a Skyfield SatelliteArray that would support, parallel to the existing `(sat - observer).at(t)`, the maneuver `(sat_array - observer).at(t)`

Regarding this proposed API, what do you envision is the return type for `(sat_array - observer).at(t)`? Will it cause confusion for users if two very different return types can result from `(something).at(t)`?

brandon-rhodes:
I would imagine the return type being exactly the same as currently for `(sat - observer).at(t)`, merely with an additional dimension in the NumPy array(s) it holds. No other differences, at least the way I'm imagining it.

glangford:
Ah, thank you for this - I didn't understand that one object (`Geometric` I guess) would end up as a container for all of the satellites.

brandon-rhodes:
No problem! As asking about a Time with, say, 85 specific dates in it returns a `Geometric` whose position has dimensions `(3, 85)`, I am imagining that 30 satellites would make the dimensions `(3, 85, 30)` for those same 85 times.

PR #404, opened 2020-07-10
use numpy to compare floats

bnavigator:
Trying to [package python-skyfield for openSUSE](https://build.opensuse.org/package/show/devel:languages:python:numeric/python-skyfield), I kept getting assertion errors during unit testing. Using `numpy.testing.assert_allclose()` instead of the `==` operator for float arrays helps.

### DIFF
--- a/skyfield/tests/test_planetarylib.py
+++ b/skyfield/tests/test_planetarylib.py
@@ -1,4 +1,5 @@
 import numpy as np
+from numpy.testing import assert_allclose
 from skyfield.api import PlanetaryConstants, T0, load
 from skyfield.constants import AU_KM, AU_M
 from skyfield.positionlib import ICRF
@@ -43,11 +44,11 @@ def test_frame_rotation_matrices():
     ]
 
     R = frame.rotation_at(ts.tdb_jd(tdb))
-    assert (R == desired_rotation).all()  # Boom.
+    assert_allclose(R, desired_rotation)
 
     R2, Rv = frame.rotation_and_rate_at(ts.tdb_jd(tdb))
-    assert (R == R2).all()
-    assert (Rv == desired_rate).all()  # Boom.
+    assert_allclose(R, R2)
+    assert_allclose(Rv, desired_rate)
 
     # Second, a moment when the angle W is more than 2500 radians.
 
@@ -93,11 +94,11 @@ def test_frame_rotation_matrices():
     frame = pc.build_frame_named('MOON_ME_DE421')
     R = frame.rotation_at(ts.tdb_jd(tdb))
     delta = abs(R - desired_rotation)
-    assert (R == desired_rotation).all()
+    assert_allclose(R, desired_rotation)
 
     R2, Rv = frame.rotation_and_rate_at(ts.tdb_jd(tdb))
-    assert (R == R2).all()
-    assert (Rv == desired_rate).all()
+    assert_allclose(R, R2)
+    assert_allclose(Rv, desired_rate)
 
 def test_rotating_vector_into_frame():
     et_seconds = 259056665.1855896
--- a/skyfield/tests/test_planetarylib.py
+++ b/skyfield/tests/test_planetarylib.py
@@ -1,8 +1,42 @@
 import numpy as np
-from numpy.testing import assert_allclose
+import os
 from skyfield.api import PlanetaryConstants, T0, load
 from skyfield.constants import AU_KM, AU_M
 from skyfield.positionlib import ICRF
+
+
+def assert_floatsareclose(a, b, atol=None):
+    """compare a list of floats as tight as possible
+
+    https://github.com/skyfielders/python-skyfield/pull/404
+
+    Parameters
+    ----------
+    a, b : array_like float
+        Arrays to compare
+    atol : float, optional
+        maximum absolute difference
+
+    Raises
+    ------
+    AssertionError
+        if the difference of any element in `a` and `b` is too large
+        or the shapes don't match
+    """
+    if not atol:
+        atol = float(os.getenv("SKYFIELD_TEST_DEFAULT_ATOL", "1e-25"))
+    assert np.shape(a) == np.shape(b)
+    difference = np.asarray(a) - np.asarray(b)
+    if not np.all(np.abs(difference) < atol):
+        imax = np.argmax(difference)
+        maxe = np.max(difference)
+        raise AssertionError("The difference between\n"
+                             "{} and \n"
+                             "{}\n"
+                             "is larger than the allowed {:.3g}.\n"
+                             "Maximum difference at element {}: {:.7g}"
+                             "".format(a, b, atol, imax, maxe))
+
 
 def test_frame_rotation_matrices():
     # To produce the following matrices:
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Alas! It was so great having those matrices match exactly. But I guess all the tweaks I did to earn the “Boom” only apply to x86_64?

Thanks for working to make Skyfield more easily available to openSUSE users!

The `assert_allclose()` routine by default does not, alas, have tight enough bounds to be a meaningful test numerically for this particular case. Also, I find that it somewhat hinders readability, as its documentation needs to be consulted to remember exactly what it does.

In most tests I do `max(abs(a - b)) < 1e-16` or somesuch to see if the biggest difference between two elements in an array are close enough, which has the advantage of being immediately readable for someone new to the code (or for me returning a few months later). Could you try pivoting this pull request to that kind of test? Try pulling in the bounds as closely as you can — chose the specific number and exponent that most closely bounds the difference, like `2e-16` or `3e-15`. (Don’t bother going to a second digit.) That best protects Skyfield in the future against regressions that might seem like clever code but that reduce its precision.

Though one danger, I suppose, for other platforms is that I might in the future tighten a test on my own system that then makes it fail on another one, like yours. I wonder if there's some way I could run Skyfield's tests on an architecture like yours, and see ahead of time if a test constraint was too tight.

bnavigator:
> The `assert_allclose()` routine by default does not, alas, have tight enough bounds to be a meaningful test numerically for this particular case. Also, I find that it somewhat hinders readability, as its documentation needs to be consulted to remember exactly what it does.

I find it very convenient and straightforward. Especially when it comes to arrays with changing scaling, `rtol` and `atol` are the way to go. And it deals with all the type castings and shapes.
There is also `assert_almost_equal` and its more special siblings, if that suits your readability more.


> In most tests I do `max(abs(a - b)) < 1e-16` or somesuch to see if the biggest difference between two elements in an array are close enough, which has the advantage of being immediately readable for someone new to the code (or for me returning a few months later). Could you try pivoting this pull request to that kind of test? Try pulling in the bounds as closely as you can — chose the specific number and exponent that most closely bounds the difference, like `2e-16` or `3e-15`. (Don’t bother going to a second digit.) That best protects Skyfield in the future against regressions that might seem like clever code but that reduce its precision.

I can do that. But that's generally not how floats work.

> 
> Though one danger, I suppose, for other platforms is that I might in the future tighten a test on my own system that then makes it fail on another one, like yours. I wonder if there's some way I could run Skyfield's tests on an architecture like yours, and see ahead of time if a test constraint was too tight.

Sure, just branch the package on OBS into your own project. You can edit the branched project to have all the platforms you want. Upload the source manually or set up a _service to checkout a specific git branch.

bnavigator:
Maximum difference for 32-bit Intel (i586):

```
[   92s] E           AssertionError: The difference between
[   92s] E           [[ 0.99942684  0.03186286  0.01143439]
[   92s] E            [-0.03382833  0.9272754   0.37284846]
[   92s] E            [ 0.00127719 -0.37302157  0.92782179]] and 
[   92s] E           [[0.9994268420493244, 0.03186286343877705, 0.011434392191818011], [-0.03382833397374688, 0.9272754001640859, 0.37284846261060917], [0.0012771890522186643, -0.37302156798771835, 0.927821792481783]]
[   92s] E           is larger than the allowed 1e-25.
[   92s] E           Maximum difference at element 4: 1.110223e-16
```

brandon-rhodes:
> But that's generally not how floats work.

Could you outline in a bit more detail the problems with experimentally chosen bounds? So far I have had a bit of success with them, in that they have prevented me from thinking I am “improving” an algorithm (“oh! look! with this tweak it runs faster!”) when in fact I am lowering its precision. But if that’s not how floats work I want to be flexible.

(Note that I only look for very high = roughly machine precision in abstract operations like this, where an input angle is becoming a rotation matrix through basic trigonometry. For end-user calculations, I definitely agree that tests need to be based on a real-world magnitude like “planet position to within 1mm” or “right ascension to within 0.1 mas”! To be honest, I often also tend do those bounds experimentally, making them as tight as possible — but I do try to target a meaningful physical measurement and not an arbitrary, physically meaningless number of digits.)

Thanks for any thoughts you can give about your objects to that approach.

bnavigator:
My comment was not about experimentally chosen bounds but about the way the two floating point numbers are compared. By subtracting one from the other, there can occur substantial [loss of significance](https://en.wikipedia.org/wiki/Loss_of_significance), which can give you a wrong impression of how close the numbers really are. 

If you really care about getting a precise match as much as possible, I think you should sacrifice perceived "readability" and rather rely on the assumption that the numerical wizards at numpy implemented the tricky details of float comparison just right in functions like [assert_array_max_ulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_max_ulp.html)

Another problem of the formula `max(abs(a - b)) < error` is, that you compare the absolute difference. But the achievable precision is relative to the exponent. So a number of 0.0001 has a smaller allowable error than say 1e6.

brandon-rhodes:
> By subtracting one from the other, there can occur substantial loss of significance, which can give you a wrong impression of how close the numbers really are.

Could you provide a clarifying example? That would help me understand how you think floats behave.

So we can compare against my own understanding of IEEE floats: I am imagining two floats that are fairly close together, and thus have the same exponent. Subtracting the two 64-bit floats will therefore ignore the exponent and, without needing to shift any digits, directly subtract the two 52-bit fractions, yielding an exact difference with zero digits of precision lost.

Yes, we see immediately an edge case: that in a small fraction of cases the nearly-adjacent floats will in fact have exponents that differ by one instead of being equal. In that case, yes, a subtraction can yield 1 digit of loss as one of the fractions is shifted by a bit to align it with the other.

But you are clearly thinking of cases where greater precision loss is possible. Could you illustrate with a pair of specific floats, so that I can see the mechanics?

(I do see that you provided a link for “loss of significance”! But all of its examples are for floats of very different magnitudes, and thus different exponents, and I cannot see how that applies to the Skyfield tests we need to adjust here?)

bnavigator:
The linked Wikipedia article explicitly talks about two floats with same magnitude. But, you are probably right about the specific test case.

The loss comes solely from the fact that `abs(a-b) < eps` does not account for the exponent part of the returned float by `abs(a-b)`. Now if you know the scaling of the reference data a priori, you can determine the necessary factor in `abs(a-b) < eps*factor` experimentally like you proposed. Or you can just use  [`assert_array_almost_equal_ulp(a, b)`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html#numpy.testing.assert_array_almost_equal_ulp) or [`assert_array_almost_equal_nulp(a, b)`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html#numpy.testing.assert_array_almost_equal_ulp)

bnavigator:
I still believe you should not chew experimentally on testing 1 or 2 remaining significant bits after repeated catastrophic cancellation, but your fix in #411 works here too.

brandon-rhodes:
> I still believe you should not chew experimentally on testing 1 or 2 remaining significant bits after repeated catastrophic cancellation…

What approach would you suggest instead? I have generally never found a good point at which to stop and say “okay, that’s enough precision,” because in an era of radio astronomy, folks seem to appreciate every last partial decimal place that can be turned from noise into accuracy; but if there’s a more principled approach to my “as many bits as I can lock in today” approach then it might make the test suite more uniform.

PR #405, opened 2020-07-10
Replace Assay by Pytest (For distribution packages)

bnavigator:
Hi,

you probably like your own testrunner Assay best.

However it is not really suited for building the package for a distribution like openSUSE, because Assay it is not availabe as a package and it does not have the integration within the build system as Pytest has.

So I decided to patch out the assay dependencies and use pytest instead. The openSUSE python maintainer [suggested](https://build.opensuse.org/request/show/819839#comment-1263815) that I should submit the patch upstream. So here it is. Feel free to use it or keep your own :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,7 @@ install:
   - "cd tmp"
   - "pip install --upgrade pip"
   - "pip install $(echo ../dist/skyfield-*.tar.gz)'[tests]'"
-  - "pip install https://github.com/brandon-rhodes/assay/archive/master.zip"
-  - "if [[ $TRAVIS_PYTHON_VERSION < 3.7 ]] ; then pip install numpy==1.11.3 scipy==1.2.2 pandas==0.23.4 unittest2 ; else pip install pandas ; fi"
+  - "if [[ $TRAVIS_PYTHON_VERSION < 3.7 ]] ; then pip install numpy==1.11.3 scipy==1.2.2 pandas==0.23.4 unittest2 pytest ; else pip install pandas ; fi"
   - "python -c 'import pandas'"
 
 script:
--- a/builders/build_novas_tests.py
+++ b/builders/build_novas_tests.py
@@ -35,6 +35,8 @@ def main():
         'Auto-generated accuracy tests vs NOVAS (see build_novas_tests.py).'
 
         from numpy import abs, array, einsum, max
+        import pytest
+
         from skyfield import (earthlib, framelib, nutationlib, positionlib,
                               precessionlib, starlib, timelib)
         from skyfield.api import Topos, load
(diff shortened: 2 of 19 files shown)


brandon-rhodes:
> you probably like your own testrunner Assay best. … So I decided to patch out the assay dependencies and use pytest instead. The openSUSE python maintainer suggested that I should submit the patch upstream.

Thanks for providing the patch, so that we have a record over here of the adjustments involved!

Alas, you are correct: I’ll be sticking with assay, which I specifically wrote at the moment when I could bear pytest no longer. Is there any chance of your package vendoring assay by quietly installing the package alongside Skyfield as part of the same package? Note that you would not have to install the command `assay`; the module can be directly invoked instead:

```
python -m assay
```

If I think of another option I will let you know — and please feel free to continue the conversation here if you have further questions or ideas. I know that boutique test runners are simply an annoyance when you’re already donating time to create a package for distribution, and, short of changing test runners, I’m interested in any other assistance I could provide.

Thanks again for letting me know!

bnavigator:
Yeah I guess you could just locally install assay in the test environment for the skyfield package to build. Providing one more tarball besides all of those data files would not matter.

The fact that assay neither has releases on Github nor a PyPI package does not help though.

From skimming the slides of your assay design talk, I got the impression that you designed it for interactive development rather than automatic builds. A Linux distro like openSUSE mandates the unit tests to check that a package built from presumably released sources works, not that some code change during development breaks functionality.

brandon-rhodes:
> The fact that assay neither has releases on Github nor a PyPI package does not help though.

I have more cycles these days for my open source projects, so I should contemplate actually releasing assay — for several years I avoided it, lest answering any questions about it took crucial time away from Skyfield. I’ll take a look at the project again and see if I can get a release up!

> you designed it for interactive development rather than automatic builds

Its lone command line switch, `--batch`, switches it out of interactive mode, turns off all the complexity designed for fast turn-around, and prints a full report of all failing tests. I use it for Skyfield’s own CI, for example:

https://github.com/skyfielders/python-skyfield/blob/72ff2c2348c6f814bab70bba0a0fc08ecd9f05a6/.travis.yml#L30

https://github.com/skyfielders/python-skyfield/blob/72ff2c2348c6f814bab70bba0a0fc08ecd9f05a6/test-code.sh#L24

My hope is that the same mechanism will serve for a build like the openSUSE build as well; let me know if it needs any tweaks.

Once I have a release of assay completed, I'll comment back here!

bnavigator:
Re #898: Remember this one

brandon-rhodes:
I did sit down last month and do a test, and running tests with this branch took around 2.5 times longer than with Assay. So I'm going to see in March if I can get Assay working again before deciding to take that severe of a penalty.

PR #410, opened 2020-07-22
Fix datetime combine

### LINKED ISSUE #409: module 'datetime' has no attribute 'combine'
https://github.com/skyfielders/python-skyfield/blob/0af8e01b4ab58375c1816e73aecc8ac55db34853/skyfield/timelib.py#L153

`combine()` is `datetime.datetime` class method

benwu95:
fix #409

### DIFF
--- a/skyfield/timelib.py
+++ b/skyfield/timelib.py
@@ -45,11 +45,11 @@ else:
         class UTC(dt.tzinfo):
             'UTC'
             zero = dt.timedelta(0)
-            def utcoffset(self, dt):
+            def utcoffset(self, _dt):
                 return self.zero
-            def tzname(self, dt):
+            def tzname(self, _dt):
                 return 'UTC'
-            def dst(self, dt):
+            def dst(self, _dt):
                 return self.zero
 
         utc = UTC()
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thanks for providing the first step towards a fix! I wound up going with my own quick patch (mentioned above):

1. The patch uses the class method, exactly as you suggest here.
2. It also adds a test.
3. It also fixes a further issue which needed tweaking: the time object needed a timezone attached. When you tested this fix, did it work on your system? Under what version of Python? I’m wondering if some versions of Python are auto-attaching time zones, if this patch solved the issue for you.
4. The name `_dt` was unhappy enough that I decided to pivot the other way and renamed the global `dt` to `dt_module`. Thanks for prodding me to solve the inconsistency. As I had written those pieces of code at different times, it was not yet bothering me that they were using `dt` for the same thing. :)

benwu95:
Thanks for the quick patch!

I've tested this fix by Python 3.7.7, and it's worked.
```python
from datetime import datetime
from workalendar.asia import Taiwan
from pytz import timezone

date = datetime.now(timezone("Asia/Taipei"))
calendar = Taiwan()
calendar.is_working_day(date)
```

PR #419, opened 2020-07-28
Remove Duplicate Function

JoshPaterson:
The function for converting from orbital elements to state vectors is duplicated between keplerlib.py and test_elementslib.py. This commit deletes the version in test_elementslib and imports the version in keplerlib.py instead.


brandon-rhodes:
Thanks! I have not yet had time to dig into possible common code between those two modules of yours, and this is a nice simplification.

PR #420, opened 2020-07-28
Update data in gravitational_parameters.py

JoshPaterson:
The source data at [ftp://ssd.jpl.nasa.gov/pub/xfr/gm_Horizons.pck](url) was updated by JPL most recently in January 2020 according to the changelog in the file.

Another question about gravitational parameters: `keplerlib.py` requires the user to explicitly specify GM, should the `OsculatingElements` object be changed to work that way too rather than using this data file?


brandon-rhodes:
> Another question about gravitational parameters: `keplerlib.py` requires the user to explicitly specify GM, should the `OsculatingElements` object be changed to work that way too rather than using this data file?

Good question!

Kind of a huge question, really.

With all sources of data, Skyfield can do one of three things:

1. Have one permanent value. For example, `de421.bsp` could always have been the "default ephemeris" and users who didn't specify anything else would have been given its results. **Problem:** Skyfield gradually would go out of date and be less precise than, say, HORIZONS, and users wouldn't have visibility into why.
2. Have one default value but upgrade it periodically. Results get gradually more and more precise as the years go on and the constants get better. Agreement with HORIZONS is achieved! **Problem:** you upgrade Skyfield and suddenly your script prints something different. Or a friend using an older Skyfield version tries your script and gets a different result. Issues are opened asking why Skyfield results are changing.
3. Always have the user specify the source of every parameter.

In general I have been opting for #3 lately. There are exceptions: the speed of light, for example, seems well enough known to just make a constant. But since GM currently seems to be an active area of research, then, yes, I think we should offer a named and dated constant for folks to use, and Skyfield can seamlessly offer better and better ones as the years go on without changing the output of old code.

I even included the name of the researcher, because they rarely seem to get credit for setting our fundamental constants :)

brandon-rhodes:
Oh! And I forgot to answer the actual question:

I suspect that this file should be phased out? As we continue to document and mainstream the Kepler and orbital element libraries, I would prefer a pivot to, say, NASA-provided constants files, if we can find them.

Oh, and:

I'm happy to see you back! Let me know if you are interested in further refactoring on your modules. If so, then I can open an issue, assigned to you, where I list some refactoring I was wanting, and you could let me know if any of it is work you're interested in tackling.

JoshPaterson:
> I'm happy to see you back!

Thanks! Yes, definitely feel free to assign me to issues, I would like to keep improving those modules!

brandon-rhodes:
@JoshPaterson — While we're on the subject of open PRs: I suspect we should close this, in favor of a future where folks load up files like `gm_Horizons.pck` when they need them, instead of having the data copied into Skyfield.

Issue #430, opened 2020-08-09
Getting ValueError: Failed to converge when loading comet list from Minor Planet Center


Bernmeister:
Following on from #428, using the ```master.zip``` from [here](https://github.com/skyfielders/python-skyfield/issues/428#issuecomment-670940243), running the code:

```
#!/usr/bin/env python3

import datetime, io, skyfield.api, skyfield.constants, skyfield.data.mpc

print( "Skyfield:", skyfield.__version__ )
latitude = -33
longitude = 151
now = datetime.datetime.strptime( "2020-08-08", "%Y-%m-%d" )
data = "    CK15A020  2015 08  1.8353  5.341055  1.000000  208.8369  258.5042  109.1696            10.5  4.0  C/2015 A2 (PANSTARRS)                                    MPC 93587"
timeScale = skyfield.api.load.timescale( builtin = True )
topos = skyfield.api.Topos( latitude_degrees = latitude, longitude_degrees = longitude )
ephemeris = skyfield.api.load( "de421.bsp" )
sun = ephemeris[ "sun" ]
earth = ephemeris[ "earth" ]
with io.BytesIO( data.encode() ) as f:
    dataframe = skyfield.data.mpc.load_comets_dataframe( f ).set_index( "designation", drop = False )
    body = sun + skyfield.data.mpc.comet_orbit( dataframe.loc[ "C/2015 A2 (PANSTARRS)" ], timeScale, skyfield.constants.GM_SUN_Pitjeva_2005_km3_s2 )
```

yields the following:

```
/usr/local/lib/python3.6/dist-packages/skyfield/data/mpc.py:207: RuntimeWarning: divide by zero encountered in double_scalars
  a = row.perihelion_distance_au / (1.0 - e)
/usr/local/lib/python3.6/dist-packages/skyfield/data/mpc.py:208: RuntimeWarning: invalid value encountered in double_scalars
  p = a * (1.0 - e*e)
/usr/local/lib/python3.6/dist-packages/skyfield/keplerlib.py:242: RuntimeWarning: invalid value encountered in double_scalars
  dE = dM/(1 - e*cos(E))
Traceback (most recent call last):
  File "magnitudeTest.py", line 17, in <module>
    body = sun + skyfield.data.mpc.comet_orbit( dataframe.loc[ "C/2015 A2 (PANSTARRS)" ], timeScale, skyfield.constants.GM_SUN_Pitjeva_2005_km3_s2 )
[...]

PR #431, opened 2020-08-12
WIP Fix for parabolic comets

JoshPaterson:
This should fix the error in #430 caused by parabolic orbits. I've marked it WIP because I still need to add tests to make sure of that.


brandon-rhodes:
@JoshPaterson — I'd be happy to help out by landing this and adding a test myself, but GitHub makes it look like you entirely replaced the Kepler file with a new one, making the diff impossible to read. Could you double-check whether you generated a whitespace change of some sort, like maybe swapping Windows line endings for the native Unix ones?

JoshPaterson:
I was struggling to figure out why that was happening, thanks for the tip!

JoshPaterson:
Also, yes, feel free to land this and add some tests!

PR #439, opened 2020-08-20
Updated: Prototype support for parallel SGP4 across multiple satellites

glangford:
This is an updated prototype to test package the parallel SGP4 capability for computing "n satellites x m times" into one routine in Skyfield.

The new function `positions_for(satellites, earthLocation, times)` takes an array of EarthSatellite objects, a `Topos` earth location and an array of `Time` and yields a position for each satellite provided. Some basic tests verify that the positions are `np.isclose()` to calculating positions one satellite at a time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ Untitled*.ipynb
 build
 htmlcov/
 stations.txt
+env/
+design/active.txt
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,4 @@ Untitled*.ipynb
 build
 htmlcov/
 stations.txt
-env/
 design/active.txt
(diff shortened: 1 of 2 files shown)


glangford:
Based on profiling for a test application, I will add some changes to improve performance. These will be around precomputing values for coordinate conversion that can be shared across satellites.

brandon-rhodes:
Why did I leave this un-merged back in 2020? Was I thinking that I should wait until it becomes something we can merge into the main code of Skyfield, and then merge?

In any case, my opinion today is that it's fine to have contributor designs like this sitting in the `designs/` folder, so I'll go ahead and merge. Thanks!

PR #475, opened 2020-11-09
Fix find_maxima call in find_minima function

Deuchnord:
Pass `epsilon`and `num` parameters from `find_minima` to `find_maxima`.

I'd like to take advantage of this PR to add documentation about these parameters and the `rough_period` attribute set to the function, since they seem very important in the calculation. Personally, I've always failed to really understand what their purpose is. Can you explain these, so I can update the docs?

### DIFF
--- a/skyfield/searchlib.py
+++ b/skyfield/searchlib.py
@@ -85,7 +85,7 @@ def find_minima(start_time, end_time, f, epsilon=1.0 / DAY_S, num=12):
     def g(t): return -f(t)
     g.rough_period = getattr(f, 'rough_period', None)
     g.step_days = getattr(f, 'step_days', None)
-    t, y = find_maxima(start_time, end_time, g, epsilon=1.0 / DAY_S, num=12)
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
I would prefer to update docs in a separate PR, to keep the history simpler, so I will be happy to go ahead and merge this one as-is! Thanks for pointing out this bug and contributing a fix.

Instead of my writing up a description here in this PR's comments about what those parameters mean, and then your having to repeat it as documentation, it might make more sense for me to just add some documentation to their docstrings directly. I was planning on writing some documentation this week (as there are a few other documentation-related issues open), so I could these parameters too.

Currently the only place I explain their meaning is in this talk, if you wanted to go watch it or review its slides:

https://rhodesmill.org/brandon/talks/#fiery-reentry

Deuchnord:
OK, then I put this PR as ready for review :)

brandon-rhodes:
Are there supposed to be CI checks that run on this repository? Did they get turned off as part of that transition that Travis CI is making? I'd better to take a look this afternoon.

PR #487, opened 2020-11-25
SpiceKernel allows slicing and tuple indexing

zanzibar7:
SpiceKernel is implemented with a dictionary instead
of integer indexing, but in user-land, it's natural
to try slices sometimes.  This allows the following
code to run ...

    from skyfield.api import load
    bodies = load('de421.bsp')
    earth, mars = bodies[3:5]
    earth, mars = bodies[('mars','earth')]

### DIFF
--- a/skyfield/jpllib.py
+++ b/skyfield/jpllib.py
@@ -164,6 +164,11 @@ class SpiceKernel(object):
 
     def __getitem__(self, target):
         """Return a vector function for computing the location of `target`."""
+        if isinstance(target, slice):
+            i = range( *target.indices(10) )
+            return ( self[x] for x in i )
+        if isinstance(target, tuple):
+            return ( self[x] for x in target )
         target = self.decode(target)
         segments = self.segments
         segment_dict = dict((segment.target, segment) for segment in segments)


brandon-rhodes:
Two concerns:

1. Since ephemerides are spare — many include only a smattering of bodies, instead of contiguous blocks of objects — I think users will most naturally expect an ephemeris to behave like a Python dictionary, not a list. And Python dictionaries don’t support slicing.
2. The syntax that results looks ambiguous. An ephemeris is an array of segments. Users will be accustomed to things like `bodies['Earth']` and may not be familiar with the SPICE integer codes. Will they think `[3:5]` means “Earth and Mars”, or mightn’t some users read it as “the 4th and 5th segments that print out when I do `print(bodies)`”? Yes, the actual array is only accessed through `.segments`. But seeing a slice up at the top-level get-item seems to me to more quickly bring to mind the file's actual content of segments, rather than body numbers.
3. What happens when users try `['Earth':'Jupiter']`?

Given the possible ambiguity, I am wary of adding this additional code and complexity. But I am open to hearing further thoughts.

zanzibar7:
1.  Well, I'm a naive user (of skyfield), and when I saw the integer-indexing, I expected slices to work -- that's just one of the standard python idioms.  Only after digging in further did I realize it was a dictionary.  So this was a bump on my way up the learning curve.  I wish there was some universal syntactic hint to help distinguish between lists and dicts.
2.  Yeah, `bodies['Earth']` is more readable -- don't know how it is that I picked up the integer indexing first -- maybe I was (needlessly) concerned about case-sensitivity.  I have no idea what you mean when you're talking about `segments` -- I was just thinking of these as reference objects for coordinate calculations.  So your alternative interpretation never came to my mind.
3.  Well, that interesting.  Right now, we get an error raised by `slices.indices` that seems reasonably informative:

> TypeError: slice indices must be integers or None or have an __index__ method

Just a suggestion.  This seems very minor, in terms of complexity growth, compared to the massive lifts with organizing time-scales, coordinate systems, and reference data, physics, ...  already accomplished.

You didn't mention the tuple-indexing -- maybe that's worth keeping even if the index-slicing isn't.

brandon-rhodes:
Thanks again for suggesting an idea—but as it makes Skyfield more complicated in a way that I'm wary of, and as no other users have commented in favor of the idea in the ensuing year and a half, I'm going to go ahead and respectfully close this pull request.

PR #495, opened 2020-12-01
Add m_per_s() to units.Velocity

### LINKED ISSUE #494: Add Velocity converstion to meters per second
I would like to propose adding a property to the `Velocity` class to report the velocity in m/s, similar to how `Distance` is available in meters.

element15:
Resolves #494


brandon-rhodes:
While I don’t want the object to grow too many reified arrays, this is perfectly symmetrical with what the Distance does, so let’s go for it!

PR #499, opened 2020-12-11
Kind of proof that this job would fail?

brunobord:
This PR is surely not mergeable. It's kind of "proof of concept"... When I tried to work on Skyfield, I've failed to found how to run tests locally. So I decided that Travis could do that for me.
When I pushed my branch, it was apparently picked up by travis.com[^1]

The jobs are taking way too long - more than 10mn - until they timed out, as you can see here:

https://travis-ci.com/github/brunobord/python-skyfield/jobs/457985299

As I've added a few `echo` commands, we can spot here the steps where it fails:

```
$ ../test-code.sh
assay command found
```

So the first "echo" is launched, and the second is never reached.
Apparently, the "culprit" is this recent change in the `test_code.sh` file:

```sh
if grep ' $' $(git ls-files design examples skyfield | grep '\.py$')
then
    echo
    echo 'Error: trailing whitespace detected on the above-listed lines'
    exit 1
fi
```

In the next commits, I'll try to add spy commands to eventually investigate, but I fear that the lack of reliable CI may block further contributions.

[^1]: https://travis-ci.com/github/brunobord/python-skyfield/ - What's important is that it's the `.com`, not the `.org`, because apparently their behaviour is not the same. I was sort of forced to migrate to the `.com` version, because:

1. the `.org` is obsolete and will be deprecated: https://docs.travis-ci.com/user/migrate/open-source-repository-migration#q-what-will-happen-to-travis-ciorg-after-december-31st-2020
2. Jobs took way too long before being launched (more than hour sometimes) and I'm patient, but there are some limits...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - "pip install --upgrade pip"
   - "pip install $(echo ../dist/skyfield-*.tar.gz)'[tests]'"
   - "pip install https://github.com/brandon-rhodes/assay/archive/master.zip"
-  - "if [[ $TRAVIS_PYTHON_VERSION < 3.7 ]] ; then pip install numpy==1.11.3 scipy==1.2.2 pandas==1.0.0 unittest2 ; else pip install pandas ; fi"
+  - "if [[ $TRAVIS_PYTHON_VERSION < 3.7 ]] ; then pip install numpy==1.11.3 scipy==1.2.2 pandas unittest2 ; else pip install pandas ; fi"
   - "python -c 'import pandas'"
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - "pip install --upgrade pip"
   - "pip install $(echo ../dist/skyfield-*.tar.gz)'[tests]'"
   - "pip install https://github.com/brandon-rhodes/assay/archive/master.zip"
-  - "if [[ $TRAVIS_PYTHON_VERSION < 3.7 ]] ; then pip install numpy==1.11.3 scipy==1.2.2 pandas unittest2 ; else pip install pandas ; fi"
+  - "pip install numpy scipy pandas unittest2"
   - "python -c 'import pandas'"
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - "pip install --upgrade pip"
   - "pip install $(echo ../dist/skyfield-*.tar.gz)'[tests]'"
   - "pip install https://github.com/brandon-rhodes/assay/archive/master.zip"
-  - "if [[ $TRAVIS_PYTHON_VERSION < 3.7 ]] ; then pip install numpy==1.11.3 scipy==1.2.2 pandas==0.23.4 unittest2 ; else pip install pandas ; fi"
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
Interesting! I had assumed the 10-minute timeouts were simply part of Travis CI .org starting to fall apart ahead of its deprecation. My hope is to soon more to GitHub Actions instead, as they seem to be the default move now that Travis CI no longer offers unlimited testing to open source projects, but I haven’t yet wanted to sit down and dedicate a day to the project.

If it’s hanging, that’s presumably because the `git` command is returning no output, so the `grep` is given no filenames and tries to grep stdin instead.

Here's an easy experiment to see if that's the case: put `< /dev/null` at the end of the `grep` command. If it's indeed getting no filenames, the `< /dev/null` should immediately end its attempt to search for matching lines in its stdin.

brunobord:
I think I've narrowed it down: https://travis-ci.com/github/brunobord/python-skyfield/jobs/458039431

At the "job time", the current directory is `/home/travis/build/brunobord/python-skyfield/tmp` and it only contains data files, no python files, so *effectively*, it returns zero python files.

Would a `cd ..; git ls-files` command work here?

I'll test!

brandon-rhodes:
The `cd ..` would break the script if the user is indeed sitting at the repo root. Instead, what about a `cd` to the root directly?

brunobord:
Oh. My. $DEITY.

I think I've found a glitch in your script.

Correct me if I'm wrong, but you're fetching:

* files in the `design`, `examples` and `skyfield` directories,
* files that and in `.py`
* and **at the same time** files that end by a space character (hence the `grep ' $'`)

If a file ends with .py, it can't ends with a space character, and vice-versa...

I've made a test by adding a file named `ab ` and it was not found by the command... Are you sure you want to keep it?

brandon-rhodes:
Not quite. The `grep ' $'` is the pattern being searched for inside the file contents themselves, not inside their filenames. You can add a stray space character to a source file to verify that the grep finds it.

brunobord:
ahhh sorry, I misread the command. New test in a few minutes.

brunobord:
Victory! https://travis-ci.com/github/brunobord/python-skyfield/builds/208083757

(but now the Python 3.6 job fails because of... doctest ; I feel lost here)

brandon-rhodes:
In travis.yml, try upgrading Pandas to 1.0.0.

brunobord:
YAY!

So I think I've managed to fix the whole job. Do you want me to:

* keep this PR here (and close it) for the "History" + open another PR with the appropriate changes, only one commit, etc?
* re-target this pull-request?

brandon-rhodes:
> So I think I've managed to fix the whole job. Do you want me to…

You can go ahead and close the pull request — I adopted its ideas piece by piece in the main build, primarily in e292756398823f0f4a257e1f7fdf7ff2a14aa56b c42c95518fe5be59c40c7752bde8942f6b4f73eb and 4ce836736ddbf16270813ca5f9b02de94ad7370e, and the build is green again.

Thanks very much for tracking down that this wasn't a Travis .org failure pending its deprecation, but an actual build problem!

brunobord:
well, I'm glad I didn't do that in vain :o)

PR #501, opened 2020-12-13
Refactor `keplerlib.stumpff`

JoshPaterson:
I was a little disappointed that this refactoring gave me a pretty much negligible speed increase.

But it's clearly a win in readability and ease of further vectorization, so I shouldn't complain too much 😀.

Also I learned a new code smell: lots of uses of `np.tile` and `np.repeat` means I'm probably broadcasting by hand instead of letting Numpy do its job.

### DIFF
--- a/skyfield/keplerlib.py
+++ b/skyfield/keplerlib.py
@@ -4,7 +4,7 @@ import sys
 import math
 from numpy import(abs, amax, amin, arange, arccos, arctan, array, cos, cosh,
                   cross, exp, log, ndarray, newaxis, ones_like, pi, power,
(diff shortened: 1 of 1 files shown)


JoshPaterson:
This is ready to merge from my end.

brandon-rhodes:
I wish I could see test suite results, like we used to before Travis CI started shutting down. I assume that the failures on the board are from earlier versions you made, and don't reflect the final slate of diffs you've submitted here?

https://travis-ci.org/github/skyfielders/python-skyfield/pull_requests

JoshPaterson:
I didn't know the CI tests were still happening, that's good to know! Turns out it was a trailing whitespace issue.

PR #511, opened 2020-12-20
Vectorized `keplerlib.propagate`

JoshPaterson:
I have successfully vectorized `keplerlib.propagate`! All of it's arguments can now be arrays. Just for testing purposes I added a function to `data/mpc.py` that can take a comets dataframe and convert it into a `_KeplerOrbit` object that represents multiple comets.

At the moment `_KeplerOrbit` objects that represent multiple orbits don't play well with other `VectorFunction` objects; I'm not sure what needs to change to fix that.

I haven't added any tests yet since none of the official functions in `data/mpc.py` can handle multiple orbits at a time yet.

Here's an example script:

```
from skyfield.api import Loader
from skyfield.data import mpc
from numpy import arange, linspace, rollaxis

load = Loader('~/projects/skyfield-data')

ts = load.timescale()

with load.open(mpc.COMET_URL) as f:
    comets = mpc.load_comets_dataframe(f)

comets_df = (comets.sort_values('reference')
          .groupby('designation', as_index=False).last()
          .set_index('designation', drop=False))

rows = comets_df.iloc[20:40]

comets = mpc._comet_orbits(rows, ts, GM_SUN)

t = ts.utc(linspace(2000, 2030, 100))
position = comets.at(t).position.au

plt.rcParams['legend.fontsize'] = 10

fig = plt.figure()
ax = fig.gca(projection='3d')

for i, (x, y, z) in enumerate(rollaxis(position, 1)):
    ax.plot(x, y, z, label=rows.designation[i])

ax.legend()

plt.show()
```
Which generates this plot:

Figure_1

### DIFF
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -228,7 +228,7 @@ def comet_orbit(row, ts, gm_km3_s2):
 
 def _comet_orbits(rows, ts, gm_km3_s2):
     e = rows.eccentricity.values
-    p = 1 - e**2
+    p = 1 - e*e
     parabolic = (e == 1.0)
     p[parabolic] = rows.perihelion_distance_au.values[parabolic] * 2.0
     p[~parabolic] *= rows.perihelion_distance_au.values[~parabolic]/ (1.0 - e[~parabolic])
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -226,7 +226,7 @@ def comet_orbit(row, ts, gm_km3_s2):
     )
     comet._rotation = inertial_frames['ECLIPJ2000'].T
     return comet
-    
+
 def comet_orbits(rows, ts, gm_km3_s2):
     p = np.empty(len(rows))
     e = rows.eccentricity.values
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -227,7 +227,7 @@ def comet_orbit(row, ts, gm_km3_s2):
     comet._rotation = inertial_frames['ECLIPJ2000'].T
     return comet
 
-def comet_orbits(rows, ts, gm_km3_s2):
+def _comet_orbits(rows, ts, gm_km3_s2):
     p = np.empty(len(rows))
     e = rows.eccentricity.values
     parabolic = (e == 1.0)
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -5,7 +5,6 @@ import io
 import re
 
 import pandas as pd
-import numpy as np
 
 from ..data.spice import inertial_frames
 from ..keplerlib import _KeplerOrbit
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -5,6 +5,7 @@ import io
 import re
 
 import pandas as pd
+import numpy as np
 
 from ..data.spice import inertial_frames
 from ..keplerlib import _KeplerOrbit
@@ -222,6 +223,31 @@ def comet_orbit(row, ts, gm_km3_s2):
         gm_km3_s2,
         10,
         row['designation'],
+    )
+    comet._rotation = inertial_frames['ECLIPJ2000'].T
+    return comet
+    
+def comet_orbits(rows, ts, gm_km3_s2):
+    p = np.empty(len(rows))
+    e = rows.eccentricity.values
+    parabolic = (e == 1.0)
+    p[parabolic] = rows.perihelion_distance_au.values[parabolic] * 2.0
+    a = rows.perihelion_distance_au.values[~parabolic]/ (1.0 - e[~parabolic])
+    p[~parabolic] = a * (1.0 - (e[~parabolic])**2)
+
+    t_perihelion = ts.tt(rows.perihelion_year.values, rows.perihelion_month.values,
+                         rows.perihelion_day.values)
+
+    comet = _KeplerOrbit._from_periapsis(
+        p,
+        e,
+        rows.inclination_degrees.values,
+        rows.longitude_of_ascending_node_degrees.values,
+        rows.argument_of_perihelion_degrees.values,
+        t_perihelion,
+        gm_km3_s2,
+        10,
+        rows['designation'],
     )
     comet._rotation = inertial_frames['ECLIPJ2000'].T
     return comet
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -228,10 +228,10 @@ def comet_orbit(row, ts, gm_km3_s2):
 
 def _comet_orbits(rows, ts, gm_km3_s2):
     e = rows.eccentricity.values
-    p = 1 - e*e
     parabolic = (e == 1.0)
-    p[parabolic] = rows.perihelion_distance_au.values[parabolic] * 2.0
-    p[~parabolic] *= rows.perihelion_distance_au.values[~parabolic]/ (1.0 - e[~parabolic])
+    p = (1 - e*e) / (1.0 - e + parabolic)
+    p[parabolic] += 2.0
+    p *= rows.perihelion_distance_au.values
 
     t_perihelion = ts.tt(rows.perihelion_year.values, rows.perihelion_month.values,
                          rows.perihelion_day.values)
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -228,12 +228,11 @@ def comet_orbit(row, ts, gm_km3_s2):
     return comet
 
 def _comet_orbits(rows, ts, gm_km3_s2):
-    p = np.empty(len(rows))
     e = rows.eccentricity.values
+    p = 1 - e**2
     parabolic = (e == 1.0)
     p[parabolic] = rows.perihelion_distance_au.values[parabolic] * 2.0
-    a = rows.perihelion_distance_au.values[~parabolic]/ (1.0 - e[~parabolic])
-    p[~parabolic] = a * (1.0 - (e[~parabolic])**2)
+    p[~parabolic] *= rows.perihelion_distance_au.values[~parabolic]/ (1.0 - e[~parabolic])
 
     t_perihelion = ts.tt(rows.perihelion_year.values, rows.perihelion_month.values,
                          rows.perihelion_day.values)
--- a/skyfield/functions.py
+++ b/skyfield/functions.py
@@ -19,16 +19,6 @@ def dots(v, u):
     """
     return (v * u).sum(axis=0)
 
-def crosses(v, u):
-    """Given one or more vectors in `v` and `u`, return their cross products.
-
-    This works whether `v` and `u` each have the shape ``(3,)``, or
-    whether they are each whole arrays of corresponding x, y, and z
-    coordinates and have shape ``(3, N)``.
-
-    """
-    return array([v[1]*u[2] - v[2]*u[1], v[2]*u[0] - v[0]*u[2], v[0]*u[1] - v[1]*u[0]])
-
 def _T(M):
     """Swap the first two dimensions of an array."""
     return rollaxis(M, 1)
--- a/skyfield/functions.py
+++ b/skyfield/functions.py
@@ -19,6 +19,16 @@ def dots(v, u):
     """
     return (v * u).sum(axis=0)
 
+def crosses(v, u):
+    """Given one or more vectors in `v` and `u`, return their cross products.
+
+    This works whether `v` and `u` each have the shape ``(3,)``, or
+    whether they are each whole arrays of corresponding x, y, and z
+    coordinates and have shape ``(3, N)``.
+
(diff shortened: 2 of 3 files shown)


skyfield/functions.py
```diff
@@ -19,6 +19,16 @@ def dots(v, u):
     """
     return (v * u).sum(axis=0)
 
+def crosses(v, u):
```
brandon-rhodes, line 22: Ah, so cross products now come to the fore! Could you try testing the code with `_cross()` from `sgp4lib.py`? I have been wanting to try it on another use-case but hadn't found a significant one yet. This could be a great chance to see if it speeds up or slows down (or breaks?) another computation besides the one use case in that library.

skyfield/data/mpc.py
```diff
@@ -226,6 +227,31 @@ def comet_orbit(row, ts, gm_km3_s2):
     comet._rotation = inertial_frames['ECLIPJ2000'].T
     return comet
 
+def comet_orbits(rows, ts, gm_km3_s2):
```
brandon-rhodes, line 230: This reads so cleanly for a vectorized routine! Could you add a `_` in front of the name for now, until we can further test it?

skyfield/keplerlib.py
```diff
@@ -578,19 +609,15 @@ def kepler(x):
...
     pcdot = -qovr0 / br * x * c1
     vcdot = 1 - bq / br * x**2 * c2
 
-    if return_1d_array:
-        position_prop = pc*position + vc*velocity
-        velocity_prop = pcdot*position + vcdot*velocity
-    else:
-        position_prop = pc*position[newaxis].T + vc*velocity[newaxis].T
-        velocity_prop = pcdot*position[newaxis].T + vcdot*velocity[newaxis].T
+    position_prop = pc[newaxis, :, :]*position[:, :, newaxis] + vc[newaxis, :, :]*velocity[:, :, newaxis]
```
brandon-rhodes, line 620: I'm happy to see `if` statements disappearing!

skyfield/data/mpc.py
```diff
@@ -226,6 +227,31 @@ def comet_orbit(row, ts, gm_km3_s2):
...
+    p = np.empty(len(rows))
+    e = rows.eccentricity.values
+    parabolic = (e == 1.0)
```
brandon-rhodes, line 233: This will be so rare that I think you could elide a step here: instead of starting with `p` empty, and then creating an array almost the size of `p` with `a * (1.0 - (e[~parabolic])**2)` only to then copy almost all those values into `p`, try just starting with `p = a * (1.0 - e**2)` which will provide you with a `p` that is mostly or entirely already populated. Then, use the `parabolic` mask to overwrite those few elements of `p` that will now be zero.

brandon-rhodes (changes requested):
Thanks for working on this! I have made a few comments but the adjustments should be fairly minor — then we'll be ready to merge. As long as the new vector-loading starts with an underscore, your tests can come later, as long as we keep passing the tests for the single-comet case!

skyfield/keplerlib.py
JoshPaterson: I could have removed that if statement a while ago thanks to `np.squeeze`!

skyfield/data/mpc.py
JoshPaterson: I might be missing something here, `a` is calculated from perihelion distance and eccentricity; if we go ahead and calculate `a` for all comets it would cause divide by 0 for any comet with e=1.0.

JoshPaterson: maybe `p` could be initialized as `1.0 - e**2` and then later multiplied by `a` for non-parabolic cases

JoshPaterson: but then we're still using a boolean mask that's almost but not quite full, and possibly reducing readability. Do you think that would still be an improvement?

JoshPaterson: It's actually not too bad after I tried it, I went ahead and committed what I came up with.

skyfield/functions.py
JoshPaterson: When I swapped that in it worked perfectly!

JoshPaterson:
I edited the script to use `enumerate` instead of `np.arange` in the for loop, which neatens things up nicely.

skyfield/data/mpc.py
```diff
@@ -226,6 +226,30 @@ def comet_orbit(row, ts, gm_km3_s2):
     comet._rotation = inertial_frames['ECLIPJ2000'].T
     return comet
 
+def _comet_orbits(rows, ts, gm_km3_s2):
+    e = rows.eccentricity.values
+    p = 1 - e**2
```
brandon-rhodes, line 231: Surprisingly, it’s generally best to square numbers — in NumPy, and floating point math more generally — by multiplying them by themselves. Try `e*e` instead :slightly_smiling_face:

skyfield/data/mpc.py
```diff
@@ -226,6 +226,30 @@ def comet_orbit(row, ts, gm_km3_s2):
...
+    parabolic = (e == 1.0)
+    p[parabolic] = rows.perihelion_distance_au.values[parabolic] * 2.0
```
brandon-rhodes, line 233: Nice, I like how you improved this section! I think you can make a final improvement: move `* rows.perihelion_distance_au.values` into the initial formula for `p`. That should avoid having to index into the big array twice with two different Boolean masks, and building those two subarrays. Then these two lines, I think, simply become `*= 2.0` and `/= 1.0 - e[~parabolic])`?

skyfield/keplerlib.py
```diff
@@ -374,9 +375,21 @@ def ele_to_vec(p, e, i, Om, w, v, mu):
...
+    low = num<end1
+    high = num>end2
+
+    num[low] = repeat(end1, sum(low, axis=1))
+    num[high] = repeat(end2, sum(high, axis=1))
+    return num
+
+
+def bracket_1d(num, end1, end2):
+    too_low = num < end1
```
brandon-rhodes, line 389: Try taking a look at `np.where()` — I think here it could eliminate the need to build 3 intermediate arrays.

brandon-rhodes (changes requested):
Oh, no! Why did GitHub not notify me about your update, and only add this to my notifications when you commented 6 hours ago? I didn't mean to leave you waiting several days. I wonder if I left the tab open and it updated it and so it thought I'd seen it?

In any case, I have two more quick comments!

brandon-rhodes:
And I'll now promptly close this tab in case leaving it open might eat notifications. Thanks for the work you're doing on this!

skyfield/data/mpc.py
JoshPaterson: I tried this but unfortunately in the parabolic case p is always initialized as 0 because 1-e**2 is zero, making the perihelion distance disappear.

Another option is this:
```
    e = rows.eccentricity.values
    p = (1 - e*e) * rows.perihelion_distance_au.values/ (1.0 - e)
    p[e==1.0] = rows.perihelion_distance_au.values[e==1.0] * 2.0
```
which works because the values are `np.float64` and any division by zero in the second line is turned into `np.inf`, which would later be replaced in the 3rd line. The downside is that this generates a warning, which we should probably avoid so users don't think something is wrong.

skyfield/keplerlib.py
JoshPaterson: This comment inspired me to discover the existence of `np.clip`, which already does what the bracket functions do. Hopefully it can replace both of them!

JoshPaterson: In newer versions of numpy this line:
```
x[past] = bracket_1d(lower[past], repeat(-bound, orb_ind), repeat(bound, orb_ind))
```
can be changed to
```
clip(lower, -bound, bound, where=past, out=x)
```
which is _so_ nice, and also means we don't need `bracket_1d` at all. Unfortunately in the version of numpy pinned in `requirements.txt` `clip` was not yet turned into a ufunc so `where` and `out` are not valid keywords. I'm assuming that means it's a no go?

JoshPaterson:
> Thanks for the work you're doing on this!

I'm glad to be able to do something useful!

skyfield/keplerlib.py
brandon-rhodes: Can you `clip()` into an intermediate array, and then do the where/out using normal assignment? (`x[past] = clipped`)

I generally keep NumPy as old as possible, to keep to a minimum the number of issues opened by folks when I accidentally use NumPy features that are too recent. I have no actual data on what versions folks are using, so if, for example, the last user of NumPy 1.11 were to shut down their old project, I would never get notified that I could advance to 1.12, so the system is not a perfect one. But I as I prefer to avoid leaving anyone behind, let’s see if we can continue forward with 1.11 for now.

skyfield/data/mpc.py
JoshPaterson: I changed this here, and in a few other places too.

skyfield/keplerlib.py
JoshPaterson: I was able to replace `bracket_1d` directly with `clip` like you suggested. 

I was also able to replace `bracket` with 2 calls to `np.copyto`, which was added in numpy 1.7 according to the documentation.

JoshPaterson:
I made a few other improvements also:
- used `np.copyto` in a number of places instead of boolean indexing
- changed `&` to `*` for boolean multiplication
- fixed a bug that came up when `t0` was a scalar

skyfield/keplerlib.py
```diff
@@ -523,74 +544,74 @@ def kepler(x):
...
                              'This is beyond the range of DT for which we '
                              'can reliably propagate states. The limits for '
                              'this GM and initial state are from {1} '
                              'to {2}.'.format(dt, kepler(-bound), kepler(bound)))
-        kfun[future] = kepler(x[future])
+        kfun[future] = kepler_1d(x[future], orb_ind)
 
-    x = amin(array([upper, amax(array([lower, (lower+upper)/2]), axis=0)]), axis=0)
+    x = copy(upper)
+    copyto(x, (upper+lower)/2, where=(lower<=upper))
```
JoshPaterson, line 579: It took me a while, but I figured out that
```upper if lower>upper else (lower+upper)/2```
is equivalent to
```min(upper, max(lower, (lower+upper)/2))```
This also occurs further down.

**Comment by JoshPaterson** [comment]
I think this is good to go now.

**Comment by xmichaelx** [comment]
Is doing something similar for mpcorb_orbit in mpc.py hard? Because that's precisely what I would need and be happy to add. 

As I understand this allows for computation of positions from multiple orbits all at once? Can this approach be used with observe method? I'd like to check whether object is in FOV of telescope for large subsets of MPC.

**Review by brandon-rhodes — COMMENTED** [review]
Drat, I apparently made this comment a few days ago but instead of submitting it, it dangled as part of a review. Let me try hitting "Comment" and that should make it appear.

File: skyfield/data/mpc.py [hunk 9]
```diff
@@ -226,6 +226,30 @@ def comet_orbit(row, ts, gm_km3_s2):
     comet._rotation = inertial_frames['ECLIPJ2000'].T
     return comet
 
+def _comet_orbits(rows, ts, gm_km3_s2):
+    e = rows.eccentricity.values
+    p = 1 - e*e
+    parabolic = (e == 1.0)
+    p[parabolic] = rows.perihelion_distance_au.values[parabolic] * 2.0
+    p[~parabolic] *= rows.perihelion_distance_au.values[~parabolic]/ (1.0 - e[~parabolic])
```
brandon-rhodes, line 234: One final idea, because it looks so painful to be creating those specially indexed copies of the Pandas columns:
```
parabolic = (e == 1.0)
p = (1 - e*e) / (1.0 - e + parabolic)
p[parabolic] += 2.0
p *= rows.perihelion_distance_au.values
```
Would that produce the same output, but without needing the special indexing?

brandon-rhodes:
@JoshPaterson, I've added the URL for this pull request to my daily to-do list so that I keep looking at it whether GitHub shows it in my notifications or not; it looks like I keep missing comments from you, and I want to get it merged by cutting down the turn-around time. I hope it hasn't made your holidays too frustrating!

skyfield/data/mpc.py
JoshPaterson: Nice! That's a clever way to avoid the division by 0. I added that in.

JoshPaterson:
> I hope it hasn't made your holidays too frustrating!

Don't worry! I know we're all volunteers, so instant feedback is nice when it happens, but I don't expect it. Also maybe I should try to get in the habit of clicking the 'request review' button so it will be sure to show up on your dashboard.

JoshPaterson:
@xmichaelx Yes, asteroids will definitely be supported once this gets officially released. The changes in this PR are mostly limited to `keplerlib.propagate`; Brandon will probably make the other necessary changes since they're in areas where he is more knowledgeable. I added the `_comet_orbits` function just so I could make that plot as an informal test.

PR #523, opened 2021-01-02
Remove `contrib/almanac2`

JoshPaterson:
I think that `skyfield.almanac` can do everything that `contrib/almanac2` can do, so there's no need to keep a copy of it in the master branch anymore. 

Probably the one part of it that would potentially be useful to someone is the vectorized version of scipy's implementation of Brent's method of finding minima, and if anyone needs that it would still be in the history.


JoshPaterson:
I should also mention that I think `skyfield.almanac` is much better than `contrib/almanac2`. It's so much shorter, more readable, and more understandable that any slight speed advantage `almanac2` has is not worth it in my opinion.

brandon-rhodes:
> I should also mention that I think skyfield.almanac is much better than contrib/almanac2. It's so much shorter, more readable, and more understandable that any slight speed advantage almanac2 has is not worth it in my opinion.

I have been keeping it around — waiting for a month when I have the time and inclination to dig back into solvers — because the speed increase was not exactly slight. If I recall, this approach, when the objective function was well-behaved, was one that you found to be “about 40% faster”. Which is substantial. It would let folks find almost twice as many events in the same amount of time? (Well, okay, that's an exaggeration; 67% more, I guess?)

brandon-rhodes:
(For my own future reference: that quote is from #211.)

PR #530, opened 2021-01-05
Fix typo in docs

JoshPaterson:


brandon-rhodes:
Thanks for catching this! Can we go with "several" instead? I generally stop updating past 3 because otherwise a specific number like this seems to spend half its life behind the true value.

brandon-rhodes:
Thanks!

PR #532, opened 2021-01-06
Vectorize `mpc.mpc_orbit`, `mpc.comet_orbit`

JoshPaterson:
This commit is only changes that are necessary to vectorize the creation of `_KelperOrbit` objects. This allows the `at()` method to work but not the `.observe` method. This PR needs to be merged before #526.

In order for the tests to pass I need to change them so that they don't use `VectorSum` objects or the `observe` method.

### DIFF
--- a/skyfield/data/mpc.py
+++ b/skyfield/data/mpc.py
@@ -5,6 +5,7 @@ import io
 import re
 
 import pandas as pd
+import numpy as np
 
 from ..data.spice import inertial_frames
 from ..keplerlib import _KeplerOrbit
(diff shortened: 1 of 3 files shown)


xmichaelx:
I see that this and #526 got stuck - is there a way I could help them go through?

xmichaelx:
Found an issue with this vectorized version, some mpcorb.dat rows cause it to generate bad results for all rows which is weird. Here is code for reproducing the issue - two generated positions are different but should be identical:

```
from skyfield.data import mpc
from skyfield.api import load, utc
import time
from io import BytesIO

text = (
    b"J95C02R 18.4   0.15 J952C  27.17980  156.55819  306.42134    7.59132  0.1449827  0.27984323   2.3148677    MPO 18139    11   1    7 days 0.75         Doppler    2000          1995 CR2           19950208\n"
    b"J95F21B  7.5   0.15 J953O   0.00000  209.62542   28.35781    0.67555  0.0000000  0.00356663  42.4258794  E MPC 25408         1    9 days              Marsden    200A          1995 FB21          19950407"
)
mpcorb_data = mpc.load_mpcorb_dataframe(BytesIO(text))

eph = load('de421.bsp')
ts = load.timescale()
t = ts.utc(2021, 10, 20)
earth = eph['earth']
sun = eph['sun']
# lets assume there is only one orbit element in result from mpcorb data
start = time.time()
orbits = mpc.mpcorb_orbit(mpcorb_data, ts, GM_SUN)
print(mpcorb_data)
start = time.time()
orbit = mpc.mpcorb_orbit(mpcorb_data.loc[0], ts, GM_SUN)
print(orbit.at(t).position.au)
print(orbits.at(t).position.au[:,0])
```

jurezakrajsek:
Hi @brandon-rhodes, @JoshPaterson 

Is there any progress on this topic of vectorizing the comet objects to enable vectorized calculations on multiple comet objects simultaneously? 
I would like to switch from the pyephem library to SkyField in my COBS project, where I currently use pyephem to calculate the visibility of comets for a given time frame and user location. 
I have approximately 3000 comets in the database, and using a for loop to calculate them one by one with SkyField is too slow for this to be usable as a real-time function on the website.

Best regards
Jure

PR #561, opened 2021-03-06
Tighten tolerances in event finding routines

### LINKED ISSUE #559: EarthSatellite.find_events() can return culmination before rise if they are close in time
When using the `find_events()` function in `EarthSatellite`, it seems a rise may be reported as culmination if the visibility is very short. This is reproduced in the following script:

```python
import skyfield.api as sf
import numpy as np
import pandas as pd

# timescale
ts = sf.load.timescale()

# location on Earth
groundstation = sf.wgs84.latlon(+53.10373, +8.85132) # NW1

# satellite
line1 = '1 00172U 19029BR  21063.59692852  .00001103  00000-0  33518-4 0  9998'
line2 = '2 00172 53.00000  36.7036 0003481 299.7327  99.3331 15.05527065  1779'
sat = sf.EarthSatellite(line1, line2, 'Starlink 172', ts)

# find rise/set
tspan = (ts.tt_jd(sat.epoch.tt-1), ts.tt_jd(sat.epoch.tt + 31))
t, events = sat.find_events(groundstation, tspan[0], tspan[1], altitude_degrees=25)

e = events[6:12]
t2 = t[6:12]
d = sat - groundstation
df =pd.DataFrame({'event': e,
    'angle': [d.at(t).altaz()[0].degrees for t in t2],
    'time': [pd.Timestamp(t.utc_iso()) for t in t2]})

print(df)
```

The output is as follows:

```
   event      angle                      time
0      0  25.002596 2021-03-03 22:38:11+00:00
1      1  68.261852 2021-03-03 22:40:30+00:00
2      2  24.995762 2021-03-03 22:42:49+00:00
3      1  25.000144 2021-03-04 00:19:38+00:00
4      2  24.999859 2021-03-04 00:19:39+00:00
5      0  25.040335 2021-03-04 19:13:42+00:00
```

The problematic event is at index 3. The exact times (`utc_jpl()`) at index 3 and 4 are 'A.D. 2021-Mar-04 00:19:38.4819 UTC' and 'A.D. 2021-Mar-04 00:19:39.0614 UTC', respectively. I think the correct behavior should be to report a rise between index 2 and 3.

This might be related to #550.

JoshPaterson:
Fixes #559.

This PR tightens tolerances as discussed in #559. It's enough to fix the particular bugs in that issue, and makes that kind of bug less likely but doesn't completely prevent it. Do you think these tolerances should be decreased more so errors are even less likely?

Ideally there would be a more thorough fix that would make those kind of bugs impossible rather than just unlikely. I'll keep thinking about it, but for now this is enough to get @bmatthiesen's scripts working.

### DIFF
--- a/skyfield/searchlib.py
+++ b/skyfield/searchlib.py
@@ -66,7 +66,7 @@ def _find_discrete(ts, jd, f, epsilon, num):
             y = y.take(indices + 1)
             # Keep only the last of several zero crossings that might
             # possibly be separated by less than epsilon.
-            mask = concatenate(((diff(ends) > 3.0 * epsilon), (True,)))
+            mask = concatenate(((diff(ends) > epsilon), (True,)))
             ends = ends[mask]
             y = y[mask]
             break
--- a/skyfield/sgp4lib.py
+++ b/skyfield/sgp4lib.py
@@ -272,7 +272,7 @@ class EarthSatellite(VectorFunction):
         doublets = repeat(concatenate(((t0.tt,), jdmax, (t1.tt,))), 2)
         jdo = (doublets[:-1] + doublets[1:]) / 2.0
 
-        trs, rs = _find_discrete(t0.ts, jdo, below_horizon_at, half_second, 8)
+        trs, rs = _find_discrete(t0.ts, jdo, below_horizon_at, half_second/4, 8)
 
         jd = concatenate((jdmax, trs.tt))
         v = concatenate((ones, rs * 2))


JoshPaterson:
This test is failing at line 68: 
https://github.com/skyfielders/python-skyfield/blob/a30a3e53a0973336ae1ac19db7d09ba0d5cb0b73/skyfield/tests/test_almanac_searches.py#L56-L68

Fixing it will have to wait until tomorrow. If it's obvious to you feel free to make whatever changes are necessary.

brandon-rhodes:
> Fixing it will have to wait until tomorrow. If it's obvious to you feel free to make whatever changes are necessary.

There's no problem with waiting til tomorrow! See what you can do about the test first; I'll probably first respond about broader issues before diving in to specific tests.

brandon-rhodes:
This was finally solved last year in the following commit:

8881e22144032d6fe6b9a71c780408c8702b79a0

I didn't notice at the time that this pull request can thereby be closed. Thanks for looking into this when it was first reported, and doing some early experiments!

PR #605, opened 2021-05-29
Fix bug in twilight calculation example script

JoshPaterson:
I found this bug when looking into discussion #597.

The example script [here](https://rhodesmill.org/skyfield/examples.html#when-will-it-get-dark-tonight) defines `previous_e` as an `ndarray` of size 1. This isn't a problem as long as the first branch of the if statement is used, but the second branch uses `previous_e` as a key to the `almanac.TWILIGHTS` dictionary which expects an integer, thus causing the error.

I was originally going to suggest using `np.isscalar` to convert the `ndarray` to an integer, but the [documentation](https://numpy.org/doc/stable/reference/generated/numpy.asscalar.html) says that function is deprecated and recommends the `item()` method instead.

### DIFF
--- a/skyfield/documentation/examples.rst
+++ b/skyfield/documentation/examples.rst
@@ -96,7 +96,7 @@ or how early I need to rise to see the morning sky:
     f = almanac.dark_twilight_day(eph, bluffton)
     times, events = almanac.find_discrete(t0, t1, f)
 
-    previous_e = f(t0)
+    previous_e = f(t0).item()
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Thanks for the fix! I'm not sure how to keep this working in the future, since there's no way to write a test for it that would exercise the opposite case where it starts mid-day rather than at midnight (without actually putting all the code in the example and making it far more complicated), but at least it will work for now for folks who start at noon.

Issue #481, opened 2020-11-22
Improvements to `keplerlib.py` and `elementslib.py`


JoshPaterson:
I'd like to start a discussion of what things in `keplerlib.py` and `elementslib.py` can be improved, and also putting them in priority order so I can work on the more important things first. Here are some ideas I've had:

- Should all of the math functions in `elementslib` have one code path that can handle floats or arrays instead of one path for arrays and one path for floats? Possibly a decorator could convert floats into arrays and then back to floats so only the array branch would be necessary
- Should the `KeplerOrbit` and `OsculatingElements` objects use au^3/d^2 to store GM internally to match the rest of Skyfield instead of km^3/s^2? I think the only place this would effect users is that `OsculatingElements` has a keyword argument `mu_km_s`. How should this be handled?
- Should the user be expected to specify a value for GM whenever it is necessary rather than saving values internally? If so how should this be handled for backwards compatibility?
- I need to refamiliarize myself with what is happening in PR #340 so it can be either merged or closed
- After the signature for `KeplerOrbit` objects is finalized it can be specified in the API reference section.
- revise the documentation for osculating orbital elements so it better matches the rest of the documentation both in tone and in the formatting of the examples

brandon-rhodes:
> I'd like to start a discussion of what things in `keplerlib.py` and `elementslib.py` can be improved…

Thanks for making some suggestions! This was a busy semester and, obviously, I didn't get around to sitting down and figuring out a list of priorities. But your list does a good job of getting ideas started, so:

> * Should all of the math functions in `elementslib` have one code path that can handle floats or arrays instead of one path for arrays and one path for floats? Possibly a decorator could convert floats into arrays and then back to floats so only the array branch would be necessary

Typically I have been able to write a single code path to support both, thanks to how NumPy lets math operators be agnostic about whether they work on scalars or strings. There are `skyfield.functions` to make things like matrix work possible without checking whether, say, time is a vector or not.

If you would like to share a sample math operation you can't see a way to make array-neutral, I can take a look and see whether it's a pattern I've been able to surmount before (or not!).

> * Should the `KeplerOrbit` and `OsculatingElements` objects use au^3/d^2 to store GM internally to match the rest of Skyfield instead of km^3/s^2? I think the only place this would effect users is that `OsculatingElements` has a keyword argument `mu_km_s`. How should this be handled?

[...]

PR #664, opened 2021-11-22
Fix Calculation of GM in `osculating_elements_of()`, etc.

JoshPaterson:
- fixes calculation of GM in `osculating_elements_of()`
- Adds calculated value for solar system barycenter to `skyfield.data.gravitational_parameters.GM_dict`. This is calculated by summing all of the masses in `GM_dict` that aren't barycenters.
- Add keyword argument to `osculating_elements_of()`. I named this argument `gm_km3_s2` to match the arguments @brandon-rhodes used in `data.mpc.py`. Most places in skyfield use `gm_km3_s2` as the argument name and I think that's the best and most explicit option. I believe `OsculatingElements.__init__()` is the last place in the public api using 'mu' in a parameter name, is there a way to deprecate this without breaking backwards compatibility?
- Make the doc string for `osculating_elements_of` more informative

### DIFF
--- a/skyfield/data/gravitational_parameters.py
+++ b/skyfield/data/gravitational_parameters.py
@@ -82,3 +82,6 @@ GM_dict = {
     2000511: 2.3312860000000000E+00,
     2000704: 2.3573170000000001E+00,
     }
+
+non_barycenters = [id for id in GM_dict.keys() if id not in range(1, 10)]
+GM_dict[0] = sum([GM_dict[id] for id in non_barycenters])
--- a/skyfield/elementslib.py
+++ b/skyfield/elementslib.py
@@ -31,9 +31,9 @@ def osculating_elements_of(position, reference_frame=None, gm_km3_s2=None):
         if position.center not in range(0, 10): # true if position.center is not a barycenter
             gm_km3_s2 += GM_dict.get(position.target, 0)
 
-    if gm_km3_s2 == 0:
-        raise ValueError("Skyfield is unable to calculate a value for GM. You"
-                " should specify one using the 'gm_km3_s2' keyword argument")
+        if gm_km3_s2 == 0:
+            raise ValueError("Skyfield is unable to calculate a value for GM. You"
+                    " should specify one using the 'gm_km3_s2' keyword argument")
 
     if reference_frame is not None:
         position_vec = Distance(reference_frame.dot(position.position.au))
--- a/skyfield/elementslib.py
+++ b/skyfield/elementslib.py
@@ -12,12 +12,19 @@ from numpy import (array, arctan2, sin, arctan, tan, inf, repeat, float64,
 def osculating_elements_of(position, reference_frame=None, gm_km3_s2=None):
     """Produce the osculating orbital elements for a position.
 
-    The ``position`` should be an :class:`~skyfield.positionlib.ICRF`
-    instance like that returned by the ``at()`` method of any Solar
-    System body, specifying a position, a velocity, and a time.  An
-    instance of :class:`~skyfield.elementslib.OsculatingElements` is
-    returned.
+    Parameters
+    ----------
+    position : :class:`~skyfield.positionlib.ICRF` object
+         commonly returned by the ``at()`` method of any Solar System body
+    reference_frame: 3x3 numpy array, optional
+        commonly used values are found in skyfield.data.spice.inertial_frames
+    gm_km3_s2: float, optional
+        Gravitational parameter (G*M) in units of km^3/s^2
+        If not specified, this is calculated for you.
 
+    Returns
+    -------
+    :class:`~skyfield.elementslib.OsculatingElements`
     """
     if not gm_km3_s2:
         gm_km3_s2 = GM_dict.get(position.center, 0)
--- a/skyfield/elementslib.py
+++ b/skyfield/elementslib.py
@@ -9,7 +9,7 @@ from numpy import (array, arctan2, sin, arctan, tan, inf, repeat, float64,
                    sinh, sqrt, arccos, arctanh, zeros_like, ones_like, divide,
                    where, pi, cross)
 
-def osculating_elements_of(position, reference_frame=None):
+def osculating_elements_of(position, reference_frame=None, gm_km3_s2=None):
     """Produce the osculating orbital elements for a position.
 
     The ``position`` should be an :class:`~skyfield.positionlib.ICRF`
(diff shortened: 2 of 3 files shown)


JoshPaterson:
@brandon-rhodes can you point me to a skyfield object whose docstring format is a good model to follow? I'm having trouble getting the doctest to pass. Thanks!

skyfield/data/gravitational_parameters.py
```diff
@@ -82,3 +82,6 @@
     2000511: 2.3312860000000000E+00,
     2000704: 2.3573170000000001E+00,
     }
+
+non_barycenters = [id for id in GM_dict.keys() if id not in range(1, 10)]
+GM_dict[0] = sum([GM_dict[id] for id in non_barycenters])
```
brandon-rhodes, line 87: Let's see if we can avoid introducing import-time side effects, like calculations. Try adding a little script in `design/` that computes this number and prints its `repr()`, then simply paste the result into this file. No point in burning the world’s fossil fuels recomputing exactly the same number every time someone imports this file, right? :wink:

(Oh: and, does this add the exact same masses that are listed in `gm_Horizons.pck`? And how does the resulting mass compare to the one used by HORIZONS?)

skyfield/elementslib.py
```diff
@@ -9,17 +9,28 @@
...
-def osculating_elements_of(position, reference_frame=None):
+def osculating_elements_of(position, reference_frame=None, gm_km3_s2=None):
     """Produce the osculating orbital elements for a position.
 
-    The ``position`` should be an :class:`~skyfield.positionlib.ICRF`
-    instance like that returned by the ``at()`` method of any Solar
-    System body, specifying a position, a velocity, and a time.  An
-    instance of :class:`~skyfield.elementslib.OsculatingElements` is
-    returned.
+    Parameters:
```
brandon-rhodes, line 15: The docstring needs to be valid Sphinx, so the errors here probably come from the erratic shifts in indentation. I simply prefer to write normal paragraphs with each line indented the same — take a look at the existing docstring here for something that Sphinx likes. Blank lines separate paragraphs.

skyfield/elementslib.py
```diff
@@ -9,17 +9,28 @@
...
+    position : :class:`~skyfield.positionlib.ICRF` object
+         commonly returned by the ``at()`` method of any Solar System body
+    reference_frame: 3x3 numpy array, optional
+        commonly used values are found in skyfield.data.spice.inertial_frames
+    gm_km3_s2: float, optional
+        Gravitational parameter (G*M) in units of km^3/s^2 If not specified, this is calculated for you.
 
+    Returns:
+    :class:`~skyfield.elementslib.OsculatingElements`
     """
```
brandon-rhodes, line 25: Remember to preserve a blank line before the terminal `"""`, per PEP-8.

skyfield/elementslib.py
```diff
@@ -9,17 +9,28 @@
                    sinh, sqrt, arccos, arctanh, zeros_like, ones_like, divide,
                    where, pi, cross)
 
-def osculating_elements_of(position, reference_frame=None):
+def osculating_elements_of(position, reference_frame=None, gm_km3_s2=None):
```
brandon-rhodes, line 12: In the docstring, specify whose GM this is — I think you mean for it to be the sum of the GM of the two bodies, but I think users might be confused and think they're only supposed to provide the GM of the body itself not the center? (And I do like the name! It's very explicit, and "GM" seems very popular among astronomy documentation that I see, which is why I used it elsewhere.)

skyfield/elementslib.py
```diff
@@ -9,17 +9,28 @@
...
+    reference_frame: 3x3 numpy array, optional
+        commonly used values are found in skyfield.data.spice.inertial_frames
+    gm_km3_s2: float, optional
+        Gravitational parameter (G*M) in units of km^3/s^2 If not specified, this is calculated for you.
 
+    Returns:
+    :class:`~skyfield.elementslib.OsculatingElements`
     """
-    mu = GM_dict.get(position.center, 0) + GM_dict.get(position.target, 0)
+    if not gm_km3_s2:
```
brandon-rhodes, line 26: Try `if … is not None:` to be a bit clearer (and faster and simpler) here.

skyfield/elementslib.py
```diff
@@ -9,17 +9,28 @@
...
+        gm_km3_s2 = GM_dict.get(position.center, 0)
```
brandon-rhodes, line 27: Put “0.0” here, which looks like a GM value, rather than “0”, which at first reading I thought was a default index into `GM_dict` (I thought you were pulling the barycenter mass, for a moment). Plus of course it saves Python from having to convert integer to float later. Also in your other `.get()`.

skyfield/elementslib.py
```diff
@@ -9,17 +9,28 @@
...
+        if position.center not in range(0, 10): # true if position.center is not a barycenter
```
brandon-rhodes, line 28: Building a `range()` object each time sounds expensive. Maybe could could build a `frozenset()` of the numbers 0 through 9 and use an `in` test against that? (Other possibilities are possible, like an `isinstance` against `int` followed by a normal inequality.)

Oh — and, instead of having a comment, try splitting this into two lines, the first of which does the test and assigns the true/false to a name like `is_barycenter`. Then no comment is needed for readers to understand the meaning.

brandon-rhodes (commented):
Thanks for the steps you're taking to address this issue, Josh! Here are a few comments before we land this and make it part of the next version of Skyfield.

brandon-rhodes:
> I believe OsculatingElements.__init__() is the last place in the public api using 'mu' in a parameter name, is there a way to deprecate this without breaking backwards compatibility?

Alas, no, I think that's now fixed. But maybe we can deprecate that class at some point? I remember our discussing whether there should be a reworking of the way that orbits relate to elements, but I don't remember what we were thinking of doing. I should return to that conversation soon when I have time.

JoshPaterson:
Oh yeah, that's right, we did discuss deprecating the OsculatingElements object in #481, hopefully I will have some time to work on that this winter!

Also, thanks for the detailed feedback, I learn a lot from you when I contribute to Skyfield, which is part of why I enjoy doing so!

skyfield/data/gravitational_parameters.py
JoshPaterson: I didn't add up all of the masses that were listed in the notes at the beginning of the file, because that would lead to most masses being double counted as individual masses and as part of the planetary barycenters. When I download the newest version of that file and add up the masses for ids >= 10 I get 132890518795.53569 km^3/s^2, and if I use horizons to calculate osculating elements for a comet with mass "n.a." orbiting the SSB it uses a slightly larger GM value of 132890518817.90877, for a difference of 22.373 km^3/s^2.

brandon-rhodes:
> Also, thanks for the detailed feedback, I learn a lot from you when I contribute to Skyfield, which is part of why I enjoy doing so!

Well, thank you for being such a good sport about it :)

So — the code introduced by this PR sets things up so that if I ask for Callisto's orbit around the Jovian barycenter, it will use the same GM as if I ask for the orbit of Jupiter around the Jovian barycenter. This is counter-intuitive to me. It feels like the orbit of Jupiter should somehow be more weighty than that of Calliso? I guess I should sit down and compute some orbits for each of them and see to what extent the result matches their real-life orbits.

Issue #796, opened 2022-09-24
reference frame from planet orientation


psbaltar:
I've worked out how to calculate a planet's orientation from the models in pck00010.tpc. So I have the RA and DEC of the planet's pole as well as its prime meridian. Now I'm trying to figure out how to build a reference frame, but it's not clear to me how to do so. How do I make a frame with the information?

I was able to calculate rotation matrices to do the conversions I needed. But I'd like to get everything into skyfield's existing framework instead of doing coordinate conversions by dealing with the matrices directly.

psbaltar:
I think I've answered my own question. After poking around, it looks like skyfield's "frames" are classes that provide a `rotation_at(t)` function that returns the rotation matrix.

brandon-rhodes:
Yes, that's indeed how you would do it! If I recall the last time I touched that code, it already works for the Moon, as documented here:

https://rhodesmill.org/skyfield/planetary.html

The implementation so far thus only understands the case where the orientation angles you mention come from a binary segment. Here's the file (that you've probably already read) where data is pulled from the text constants file to set up the frame that uses the binary segment:

https://github.com/skyfielders/python-skyfield/blob/master/skyfield/planetarylib.py

And here are comments on previous open issues discussing how a new frame type can be written for simpler bodies, where instead of a binary segment there are just angles and rates directly in the text file:

https://github.com/skyfielders/python-skyfield/issues/417#issuecomment-664385186

https://github.com/skyfielders/python-skyfield/issues/709#issuecomment-1043825160

Issue #762 might also be resolved if we get this implemented, or at least I think it would be a step closer?

Alas, I haven't had time recently to sit down and add a new feature that I myself don't need right now—but might have more time coming up this winter to contribute more often. If you're able to get something working, let me know, I'd be happy to look at a Pull Request or even just a working script pasted into an issue!

psbaltar:
After thinking about it more, I realized that generalizing planetary reference frames is more complicated than I expected.

I've been calculating ecliptic frames for other planets. Mainly to make those cool epicycle plots, but for observers on other planets. However, an ecliptic frame is only one possible frame. I would also venture to guess that it's the least interesting one for most people. Topographic frames and equatorial frames should also be available. So the question is, what would the API look like for requesting different frames for the same body?

In the meantime, I'll clean up the code I have for getting a planet's orientation and submit a PR for that.

Issue #709, opened 2022-02-16
How to set up observations from bodies other than Earth?


AJMansfield:
I'm interested in plotting what sorts of eclipses and transits would be observed from the surface of Mars and from some Jovian moons, however I'm having trouble figuring out how to set up geographic positions on bodies other than Earth.

Setting up an reference ellipsoid for an alternate body isn't a problem; one can just construct [`skyfield.toposlib.Geoid`](https://github.com/skyfielders/python-skyfield/blob/master/skyfield/toposlib.py#L138) with the appropriate parameters. E.g. for Mars:

```
# from https://ode.rsl.wustl.edu/mars/pagehelp/Content/Frequently_Asked_Questions/Coordinate_System.htm
# equatorial radius = 3396.19 km
# polar radius = 3376.2 km
iau2000_mars = Geoid("IAU2000 Mars", 3396190.0, 169.89)
```

Setting up the planet's rotation rate is a little less clear, but it seems like I'd just need to copy [`skyfield.toposlib.GeographicPosition`](https://github.com/skyfielders/python-skyfield/blob/master/skyfield/toposlib.py#L51), and replace the parameters in `lst_hours_at` with appropriate parameters for another body.

It's not all that obvious what the different constants in that function represent though, making it difficult to search for the equivalent parameters for other bodies of interest. But at least as a rough first approximation for Mars, it seems like I could just scale the output by the ratio of the mars and earth sidereal day lengths, and get at least the right average rotation rate:

```
class MarsGeographicPosition(GeographicPosition):
    ...
    def lst_hours_at(self, t):
        mars_earth_ratio = 1.0287656 # https://www.wolframalpha.com/input?i=mars+rotation+period+%2F+earth+rotation+period
        sprime = -47.0e-6 * (t.whole - T0 + t.tdb_fraction) / 36525.0
        return (t.gast + self.longitude._hours + sprime / 54000.0) * mars_earth_ratio % 24.0
    ...
```

I'd definitely like to set it up more accurately than that though.

And it's not clear at all to me how to set up obliquity parameters. Mars at least, seems to have this included in its `de421.bsp` the same as Earth, but I can't seem to find a way to set it up manually for bodies I'm only defining with e.g. a Kepler orbit.

To summarize: 

**How do I set up the parameters in `GeographicPosition.lst_hours_at` for bodies other than Earth?**
**How do I set up obliquity for e.g. Kepler Orbit bodies?**

brandon-rhodes:
[...]

Issue #417, opened 2020-07-26
Central meridian longitude of planets and Jovian satellites?


davetyp:
This is tangentially related to issue # 272.  How can one extract the central meridian longitude (CML) (the body-centric longitude pointing at Earth) for a planet or a Jovian satellite?  The information should be in de430 for the planets and jup310 for the Jovian satellites.  What call do we make to skyfield to obtain this information?  Note that for the gas giants, Jupiter in specific, one normally needs CML of System III, which is the planetary magnetosphere, not the CML of system I or II (low latitude and high latitude clouds).  These three CML Systems rotate at different speeds.

This is related to issue 272 because eight Jovian satellites are tidally locked.  That means if you know the CML of, e.g., Io, you know its position in orbit around Jupiter after you make a tiny adjustment for parallax.  This of course doesn't help for any of the Jovian satellites that are not tidally locked.

So, how do we get CML out of Skyfield?

brandon-rhodes:
> The information should be in de430 for the planets and jup310 for the Jovian satellites.

At least as delivered in `.bsp` files, those ephemerides list no orientation data, alas:

```
$ python -m jplephem spk de430t.bsp
File type DAF/SPK and format LTL-IEEE with 15 segments:
2287184.50..2688976.50  Type 2  Venus Barycenter (2) -> Venus (299)
2287184.50..2688976.50  Type 2  Mercury Barycenter (1) -> Mercury (199)
2287184.50..2688976.50  Type 2  Earth Barycenter (3) -> Earth (399)
2287184.50..2688976.50  Type 2  Earth Barycenter (3) -> Moon (301)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Sun (10)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Pluto Barycenter (9)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Neptune Barycenter (8)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Uranus Barycenter (7)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Saturn Barycenter (6)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Jupiter Barycenter (5)
2287184.50..2688976.50  Type 2  Solar System Barycenter (0) -> Mars Barycenter (4)
[...]

Issue #46, opened 2015-04-10
Geocentric ICRF position of given latitude/longitude (feature request)


barrycarter:
Brandon, not sure if you want to add this to skyfield, but it might be
useful to know the geocentric position of a given observing site
(latitude/longitude/elevation) to high precision for a given
time. This would be an alternate way of compensating for
precession. Thoughts:

  - The position of the North Pole at the J2000 epoch is, of course,
  (0,0,epr) by definition, where epr is the Earth's polar radius
  (6356.752 km per HORIZONS)

  - I used HORIZONS to find the position of the North Pole from 9999
  BC to 9999 AD. HORIZONS won't let you compute the North Pole from
  the geocenter, but you can do the opposite: compute the geocenter
  from the north Pole (so the following values should be negated):

https://github.com/barrycarter/bcapps/blob/master/ASTRO/north-pole-from-geocenter.txt.bz2

  - As expected, all 3 coordinates (x,y,z) follow a sinusoidal pattern
  over time with a period matching the Earth's precession. I
  considered retrofitting the data to sine waves, but I'm sure you can
  use a more accurate precession model.

  - The position of non-polar locations of course changes as the Earth
  rotates, and repeats every sidereal day. However, it should follow a
  reasonably simple formula, even allowing for the Earth's ellipticity
  and precession tilt.

  - You may or may not find other helpful files at:

https://github.com/barrycarter/bcapps/tree/master/ASTRO/

brandon-rhodes:
Happily, this feature is already built into Skyfield, along with the nutation and precession models used for the Astronomical Almanac. The results should agree very closely to those of HORIZONS, at least within historical time frames. If you run the following code in an IPython Notebook, do you see a plot similar to the one from HORIZONS?

```
%pylab inline
```

```
from skyfield.api import earth
years = range(-9999, 9999, 10)
north_pole = earth.topos('90 N', '0 W')
x, y, z = north_pole.gcrs(utc=(years, 1, 1)).position.km
plot(x, y)
```

This is probably a question that other people will have, so I will leave this issue open until I have time to add a section on Earth orientation to Skyfield's documentation — thanks for pointing out that people will want to know its precession model!

barrycarter:
Still playing around with this: it matches HORIZONS well between ~5000 BC and 9999 AD, but seems to breakdown a bit before 5000BC. Example:

```
from skyfield.api import *
north_pole = earth.topos('90 N', '0 W')
[...]

PR #798, opened 2022-09-24
Planetary orientation models from text PCK file

psbaltar:
This adds a `PlanetaryOrientation` class to planetarylib as mentioned in #796. The class is able to provide the planet's orientation (RA and DEC of the pole and the prime meridian), as well as the planet's shape (three radii for an ellipsoid).

The models are read from a text PCK file. There's notes about how binary PCK files should be preferred if available as they are more precise. In fact, the Earth orientation model in the text PCK file doesn't include nutation. From what I can tell, there's binary files only for Earth and Earth's moon. 

```
from skyfield.api import load, PlanetaryConstants, PlanetaryOrientation

ts = load.timescale()
t = ts.utc(2020,1,1)

pc = PlanetaryConstants()
pc.read_text(load('pck00010.tpc'))

po = PlanetaryOrientation(pc, '399')
print(po.orientation_model, po.nutation_model, po.shape_model) # True False True

po.at(t)
print(po.orientation) # [-0.12819123924090417, 89.8886076127033, 2637009.9229112]
print(po.radii) # [6378.1366, 6378.1366, 6356.7519]
```

The models are described in detail in
Archinal, B.A., A’Hearn, M.F., Bowell, E. et al. Report of the IAU Working Group on Cartographic Coordinates and Rotational Elements: 2009. Celest Mech Dyn Astr 109, 101–135 (2011). https://doi.org/10.1007/s10569-010-9320-4
Also available here: https://aa.usno.navy.mil/downloads/reports/Archinaletal2011a.pdf

A SPICE PCK file that `PlanetaryOrientation` reads the models from is available here:
https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00010.tpc

### DIFF
--- a/skyfield/api.py
+++ b/skyfield/api.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from .constants import B1950, T0, pi, tau
 from .constellationlib import load_constellation_map, load_constellation_names
 from .iokit import Loader, load_file
-from .planetarylib import PlanetaryConstants
+from .planetarylib import PlanetaryConstants, PlanetaryOrientation
 from .positionlib import position_from_radec, position_of_radec
 from .starlib import Star
 from .sgp4lib import EarthSatellite
@@ -27,7 +27,8 @@ S = W = -1.0
 __all__ = [
     'Angle', 'B1950', 'Distance', 'E', 'EarthSatellite',
     'GREGORIAN_START', 'GREGORIAN_START_ENGLAND',
-    'Loader', 'PlanetaryConstants', 'N', 'S', 'Star', 'W',
+    'Loader', 'PlanetaryConstants', 'PlanetaryOrientation',
+    'N', 'S', 'Star', 'W',
     'T0', 'Time', 'Timescale', 'Topos', 'Velocity',
     'datetime', 'iers2010', 'load', 'load_constellation_map',
     'load_constellation_names', 'load_file',
--- a/skyfield/constants.py
+++ b/skyfield/constants.py
@@ -28,5 +28,6 @@ GS = 1.32712440017987e+20
 # Time.
 T0 = 2451545.0
 B1950 = 2433282.4235
+D_JC = 36525 # days in a Julian century
 
 C_AUDAY = C * DAY_S / AU_M
--- a/skyfield/planetarylib.py
+++ b/skyfield/planetarylib.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 """Open a BPC file, read its angles, and produce rotation matrices."""
 
-from numpy import array, cos, nan, sin
+from numpy import array, cos, nan, sin, deg2rad
 from jplephem.pck import DAF, PCK
-from .constants import ASEC2RAD, AU_KM, DAY_S, tau
+from .constants import ASEC2RAD, AU_KM, DAY_S, tau, T0, D_JC
 from .data import text_pck
 from .functions import _T, mxv, mxm, mxmxm, rot_x, rot_y, rot_z
 from .units import Angle, Distance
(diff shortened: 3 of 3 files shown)


brandon-rhodes:
Thanks very much for putting together this code! Having Skyfield be able to compute and return these values will be very useful. I'll see if I can write up a quick code review this week; if not, then look for something from me by mid-October.

psbaltar:
I made some updates to vectorize the calculations. I hadn't been using Skyfield's vectorization features because I wasn't very familiar with the topic. It's something I hadn't worked with before in numpy.

I also wanted to take the opportunity to thank you for this great library! I'm not a professional astronomer nor a professional programmer, so I've learned a lot while working with Skyfield.

billyauhk:
I should point out that for Earth, Skyfield should already supports using Earth Orientation Parameters; for Moon, libration angles from an ephemeris should be used. The PCK is not necessary for Earth/Moon, and might not be the currently-accepted values.

Since WGCCRE writes a new report every 3 years, values from a PCK may not be updated to the current standard at https://astrogeology.usgs.gov/groups/IAU-WGCCRE
The 2015 standard is here: https://astropedia.astrogeology.usgs.gov/download/Docs/WGCCRE/WGCCRE2015reprint.pdf

(P.S. Not a Skyfield user on my own, but found this when I am trying to check if Skyfield could compute physical ephemerides of planets.)

Cross-ref #417 #709 as both will involve using WGCCRE values at some point.

psbaltar:
> Since WGCCRE writes a new report every 3 years, values from a PCK may not be updated to the current standard at https://astrogeology.usgs.gov/groups/IAU-WGCCRE The 2015 standard is here: https://astropedia.astrogeology.usgs.gov/download/Docs/WGCCRE/WGCCRE2015reprint.pdf

I've been using `pck00010.tcp` for my own purposes (non-research) because it's the latest I can find from NAIF. It's dated 2011-10-21 and says it's based on the 2009 WGCCRE report. I had come across the 2015 report, but I haven't gone through the exercise of checking what changed. However, I did see that they removed the orientation model for Earth because "accuracy was poor, and the expressions failed near the fundamental epoch (J2000.0), yet they were sometimes used as a recommended model".

Since the PCK file is plain text, it wouldn't be hard for a user to make any necessary updates if they wanted to.

billyauhk:
For Earth/Moon, the models are kind-of *removed* from the 2015 report in order to prevent people misusing them and treat it as the "correct" expressions. Precession-nutation models for Earth are already inside Skyfield but one may need some additional code to retrieve what you want (I don't know what you want at the very end).

Some variation of the code at https://github.com/skyfielders/python-skyfield/issues/46#issuecomment-95299079 will be useful in deriving a rotation matrix (and thus a milli-arcsecond precise orientation of the Earth), so adding new code with known and inferior precision and add confusion to users sounds wary to me.

@psbaltar I don't think "my own purposes (non-research)" is a responsible way to add code into a public repository, without highlighting the possible side-effect which may affect others. Known problems/deficiencies should at least be when documented (this is why I added my opinion in the first place).
I have no opinion on using a PCK for any other Solar System Bodies other than Earth/Moon

psbaltar:
@billyauhk Sorry for the misunderstanding. I thought that the warnings against using the PCK files were meant as a blanket statement to not use them because they may be out of date.

I wasn't trying to imply/recommend that the PCK data should be used for any particular cases. This is just provided as generic code that's capable of reading a model from a given file. While there are already better models for Earth/Moon, this is a way to get models for the other planets. The data file it is intended to read just happens to include Earth and Moon models as well, even though they aren't current/best. In fact, the PCK file itself cautions against using its Earth rotation model.

@brandon-rhodes would have to weigh in on how best to make the user aware of the limitations.

brandon-rhodes:
@psbaltar — My apologies for taking so long to respond to this pull request! The last few months have been very busy, but I'm returning to my GitHub notifications at last.

First, I want to reassure you that I don't find this contribution at all irresponsible! It’s not as though you are adding new pages to the Skyfield documentation, telling users to ignore the normal `wgs84.latlon()` and `ITRS` techniques for Earth orientation and replacing them with a PCK file. When this feature is documented, we can simply highlight its use in things like the orientation of Mars and Jupiter and Saturn, and steer people away from old PCK files that might list the Earth.

Second, I wanted to ask if you're interested in my writing up a full code review—or whether I have delayed too long in answering, and by now you've lost interest and moved on to other projects? If you're still around, I would be happy to make comments on the code—just let me know either way, thanks!

psbaltar:
@brandon-rhodes Thanks for getting back to me. A code review would be great. Planetary imaging season is over, so I've back burnered this. I do plan on circling back to it though.

Either way, I already have the code and would like to contribute to Skyfield since it's been so helpful!

brandon-rhodes:
Thanks for the reply! The next couple of weeks will have me busy with preparations for PyCon, but I'll do the code review right after the conference if I don't get to it beforehand.

PR #824, opened 2023-01-06
Load deepsky objects from pyongc

### LINKED ISSUE #823: NGC and IC objects integration
I think it would be nice to have deep sky objects integration in skyfield, like hipparcos database.
I have been maintaining OpenNGC database and it's python interface PyOngc for some time and I wonder if you're willing to consider a PR about using PyOngc as data provider for such objects. I think I can reuse most of the code of the Star class by moving it under a parent `DistantObject` class, then override specific code for Star and DeepSky.

mattiaverga:
Loading deepsky objects from pyongc is possible in two ways: a simple load using `Star.from_ongc('identifier')` or by converting a dataframe from pyongc.data with `skyfield.data.ongc.load_dataframe()` + `Star.from_dataframe()` for loading several objects at once.
The pyongc dependency is optional (users needs to install skyfield[ongc] extra to pull in pyongc dependency).

Fixes #823

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,4 +60,7 @@ setup(
         'numpy',
         'sgp4>=2.2',
         ],
+    extras_require={
+        'ongc': ['pyongc'],
+    },
 )
--- a/skyfield/data/ongc.py
+++ b/skyfield/data/ongc.py
@@ -1,0 +1,29 @@
+from math import pi
+
+PANDAS_MESSAGE = """Skyfield needs Pandas to load the OpenNGC catalog
+
+To load the OpenNGC star catalog, Skyfield needs the Pandas data
+analysis toolkit.  Try installing it using your usual Python package
+installer, like "pip install pandas" or "conda install pandas".
+"""
+
+def load_dataframe(in_dataframe):
+    """Convert a dataframe from pyongc.data for import in skyfield."""
+    try:
+        from pandas import read_csv
+    except ImportError:
+        raise ImportError(PANDAS_MESSAGE)
+
+    df = in_dataframe[['name', 'ra', 'dec', 'parallax', 'pmra', 'pmdec']].copy()
+
+    df = df.assign(
+        ra_hours = df['ra'] * 180 / pi / 15.0,
+        dec_degrees = df['dec'] * 180 / pi,
+        epoch_year = 2000,
+    )
+
+    df.rename(columns = {'parallax':'parallax_mas',
+                         'pmra':'ra_mas_per_year',
+                         'pmdec':'dec_mas_per_year'}, inplace = True)
+
+    return df.set_index('name')
--- a/skyfield/starlib.py
+++ b/skyfield/starlib.py
@@ -106,6 +106,18 @@ class Star(object):
             epoch=epoch,
         )
 
+    @classmethod
+    def from_ongc(cls, identifier):
+        from pyongc import ongc
+        deepsky = ongc.Dso(identifier)
+        return cls(
+            ra = Angle(radians=deepsky.rad_coords[0]),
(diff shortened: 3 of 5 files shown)


setup.py
```diff
@@ -60,4 +60,7 @@ def make_distribution(self):
         'numpy',
         'sgp4>=2.2',
         ],
+    extras_require={
```
brandon-rhodes, line 63: For now let's avoid changing `setup.py`. If we tell users in the documentation that they need to install Pandas and `pyongc`, then I think that will be enough to get them going if their project needs NGC objects.

skyfield/starlib.py
```diff
@@ -106,6 +106,18 @@ def from_dataframe(cls, df):
             epoch=epoch,
         )
 
+    @classmethod
```
brandon-rhodes, line 109: Let's avoid adding any new methods to `Star`, since it will become a very large class if it has to learn details about every kind of external catalog. Instead, it is designed to stay simple, and every catalog simply needs to format its Pandas dataframe for the `from_dataframe()` method, and then the star-catalog details remain over in the code specific to those catalogs, instead of needing updates to the class itself, which ideally shouldn't even need to know that different catalogs exist.

skyfield/tests/test_stars.py
```diff
@@ -1,8 +1,25 @@
...
+from skyfield.data import hipparcos, ongc
 
 def test_dataframe():
     with api.load.open('hip_main.dat.gz') as f:
-        df = load_dataframe(f)
+        df = hipparcos.load_dataframe(f)
     star = api.Star.from_dataframe(df)
     assert repr(star) == 'Star(ra shape=9933, dec shape=9933, ra_mas_per_year shape=9933, dec_mas_per_year shape=9933, parallax_mas shape=9933, epoch shape=9933)'
+
+def test_dataframe_from_ongc():
```
brandon-rhodes, line 11: Thank you for including tests! Many developers don't think to write them, and it's very helpful.

skyfield/tests/test_stars.py
```diff
@@ -1,8 +1,25 @@
...
+    ongc_data = data.all()
+    df = ongc.load_dataframe(ongc_data)
+    hercules = api.Star.from_dataframe(df.loc['NGC6205'])
+    assert repr(hercules) == (
+        'Star(ra=250.42345833333337, dec=36.46130555555556, ra_mas_per_year=-3.18, '
```
brandon-rhodes, line 16: Normally I would object to a `repr()` since even very small changes in how numbers are input and rounded will break the tests, but in this case it's a bit elegant—it lets us very quickly verify that all the essential properties of the star are correct. So let's keep it this way, and see if it ever breaks!

skyfield/data/ongc.py
```diff
@@ -0,0 +1,29 @@
...
+To load the OpenNGC star catalog, Skyfield needs the Pandas data
+analysis toolkit.  Try installing it using your usual Python package
+installer, like "pip install pandas" or "conda install pandas".
+"""
+
+def load_dataframe(in_dataframe):
+    """Convert a dataframe from pyongc.data for import in skyfield."""
+    try:
+        from pandas import read_csv
+    except ImportError:
```
brandon-rhodes, line 14: You did a good job repeating the patterns in `hipparcos.py`!

brandon-rhodes (changes requested):
Thanks for writing this up—I've asked for one or two tweaks, but overall the code looks solid, and like a contribution that will help folks get coordinates for galaxies!

mattiaverga:
I've pushed the requested changes, let me know if it's now ok.

Issue #873, opened 2023-07-08
A bug in nutationlib's equation_of_the_equinoxes_complimentary_terms()


trufanov-nok:
I was trying to debug my TEME<->ITRF conversion by comparing to skyfield results and found out a point there my results are starting to differ. It's here: https://github.com/skyfielders/python-skyfield/blob/master/skyfield/nutationlib.py#L177  

```
    fa[4] = (( 450160.398036 +
              (-482890.5431 +
              (      7.4722 +
              (      0.007702 +
              (     -0.00005939)
              * t) * t) * t) * t) * ASEC2RAD
              + (-5.0*t % 1.0) * tau)
```
In particular at `-5.0*t % 1.0`.   
I was surprised that modulo operation with negative value behaves weird in python. For example:
```
>>> t =  0.16322491910627326
>>> -5.0*t % 1.0
0.1838754044686337
```
while I would expect 
```
>>> -5.0*t 
-0.8161245955313663
```

And that's how it would work in C with `fmod(-5.0*0.16322491910627326, 1.0)`.

So I tried to find an article or another source for this method. And in SOFA's C codebase (http://iausofa.org/2021_0512_C/sofa_c-20210512.tar.gz) at `nut80.c::224` I found:

```
/* Longitude of the mean ascending node of the lunar orbit on the */
/* ecliptic, measured from the mean equinox of date. */
   om = iauAnpm(
        (450160.280 + (-482890.539 + (7.455 + 0.008 * t) * t) * t)
        * DAS2R + fmod(-5.0 * t, 1.0) * D2PI);
```

I assume that C code was a source and then it was ported to python. In this case Skyfield has a bug because C `fmod(-5.0 * t, 1.0)` is not equivalent to python's `-5.0*t % 1.0`.    
I don't know python lang but perhaps it should be?:
```
>>> 5.0*t % 1.0 * -1
-0.8161245955313663
```

trufanov-nok:
Or better, keeping in mind that the module has `from numpy import ... fmod` to use `fmod(-5.0*t, 1.0)`   
Also the `fa $= tau` at https://github.com/skyfielders/python-skyfield/blob/master/skyfield/nutationlib.py#L189 will produce incorrect results. It should be replaced with `fa = fmod(fa, tau)`

brandon-rhodes:
[...]

PR #874, opened 2023-07-09
Fix modulo op for negative values

trufanov-nok:
use numpy's fmod instead of python's % to be equal to C fmod() in case of negative values.
resolves https://github.com/skyfielders/python-skyfield/issues/873

### DIFF
--- a/skyfield/nutationlib.py
+++ b/skyfield/nutationlib.py
@@ -174,7 +174,7 @@ def equation_of_the_equinoxes_complimentary_terms(jd_tt):
               (      0.007702 +
               (     -0.00005939)
               * t) * t) * t) * t) * ASEC2RAD
-              + (-5.0*t % 1.0) * tau)
+              + fmod(-5.0*t, 1.0) * tau)
 
     fa[ 5] = (4.402608842 + 2608.7903141574 * t)
     fa[ 6] = (3.176146697 + 1021.3285546211 * t)
@@ -186,7 +186,7 @@ def equation_of_the_equinoxes_complimentary_terms(jd_tt):
     fa[12] = (5.311886287 +    3.8133035638 * t)
     fa[13] = (0.024381750 +    0.00000538691 * t) * t
 
-    fa %= tau
+    fa = fmod(fa, tau)
 
     # Evaluate the complementary terms.
 


brandon-rhodes:
@trufanov-nok — Thanks for pointing this out! I will also want to write a test that prevents a contributor from breaking this again in the future. Could you share the specific coordinates that led you to discover the difference in how modulo was handled, so that I could use them in the test? Thanks!

trufanov-nok:
The test case would be
```
from skyfield import nutationlib
nutationlib.equation_of_the_equinoxes_complimentary_terms(jd_tt=2457506.7901703566)
```
It's a date somewhere in 2016 if I racall right   
It'll give you after patch a `fa` array:
[ 4.63751774e+00  8.26363095e+00  2.16215241e+00  1.07606250e+01
 -3.32755093e+00  4.30222197e+02  1.69882417e+02  1.04308925e+02
  6.07306002e+01  9.24542295e+00  4.35558967e+00  6.70191591e+00
  5.93431245e+00  3.97985269e-03]

There the `fa[4]` is negative and it's the only value that will be counted in `a = ke1.dot(fa)` because `ke1 = [0 0 0 0 1 0 0 0 0 0 0 0 0 0]`.

brandon-rhodes:
> It's a date somewhere in 2016 if I recall right
> It'll give you after patch a fa array…

I get only a scalar value:

    2.1371671320680886e-09

Why would the output from the `equation_of_the_equinoxes_complimentary_terms(jd_tt=2457506.7901703566)` switch to an array, when given a scalar input like `2457506.7901703566`? Maybe I misunderstood your comment?

trufanov-nok:
When I referred to `fa` array and `a` variable I meant local variables which values I can check via `print` command. The function result is scalar.

brandon-rhodes:
@trufanov-nok — Thank you for explaining! I understand now that you were referring to the internal value `fa` that's inside the function. But for a test case, I'm going to need a value that's visible outside the function, because the only thing a test case can do is call `equation_of_the_equinoxes_complimentary_terms()` and look at the result. Do you have a date I can try calling it with, that will give a different result once your patch is applied? With the two changes in your patch applied, I still get `2.137167132068086e-09` as the return value. (But maybe I've applied them wrong, so please double-check me!)

(Also: could you double-check whether the first line you change in your patch makes a difference to the numeric outcome? I cannot tell quite why, in the stack of equations for `fa[0]` and `fa[1]` and so forth, you choose `fa[4]` alone to adjust.)

trufanov-nok:
The `fa[4]` is chosen because python's modulo operation behaves differently only with negative arguments, and assuming time `t` is always positive the `fa[4]` is the only place that have a negative coefficient.    

Indeed, I was comparing the local variables instead of function result. And now I see that effect on final result is almost neglectable small:
for t = 2457506.7901703566
before patch: 2.1371671320680886e-09
after patch:    2.137167132068086e-09

brandon-rhodes:
```
before patch: 2.1371671320680886e-09
after patch:  2.137167132068086e-09
```

Yes, that difference looks to be down at the level of the machine precision of a 64-bit floating point number. Unless this affects an astronomical computation to within its error bounds — and please re-open this PR if you find a case where it does! — I am inclined to keep using the built-in `%` Python operator, whose operation is much simpler to reason about — since something like '% 7.0' has a simple-to-explain output range [0.0,7.0), instead of fmod's rather crazy range (-7.0,7.0) where pairs of values like -4.0 and +3.0 kind-of mean exactly the same thing except expressed differently.

PR #909, opened 2023-10-15
Add Tycho2 catalog

wheirman:


brandon-rhodes:
Thank you for this contribution! Could you add a quick test for Skyfield's test suite, so that the feature will keep working in the future without other folks likely to break it? You could add it right next to the hipparcos test:

https://github.com/skyfielders/python-skyfield/blob/41b44cf2a5d18fa963a5db1bcf27e413865fd3ce/skyfield/tests/test_stars.py#L4

With a passing test this will be ready to merge. Thank you again!

wheirman:
> Could you add a quick test for Skyfield's test suite, so that the feature will keep working in the future without other folks likely to break it?

I added a test. Looks like the hipparcos test includes a full copy of `hip_main.dat`. `tyc_main.dat` is 344MB so I don't think including that in the Git repo is a good thing, nor is downloading it on each test. So I included just the first 10 lines as `tyc_main_head.dat` which should be good enough to test the functionality of the importer.

brandon-rhodes (changes requested):
Thanks for putting this together! Two quick questions that I'm unsure of the answers to:

1. Does `epoch_year = 1991.25` need to be changed to `2000.0`? The Tycho docs almost make it sound like it uses a different epoch year than Hipparcos did. Maybe we need to generate a sample star location for a fast-moving star, like Polaris, and verify that its position isn't 8.75 years off? Once this lands, I should add a test for Hipparcos that does the same thing.
2. Have you verified that each of the `na_values` is actually used in Tycho? For Hipparcos, they had to be hand-selected so that columns would be read in as numbers rather than as a huge number of separate strings. (When printing the `df.info()`, I could see that what should have been floating-point fields were coming in as strings, because Pandas didn't know that all-spaces mean the number NaN.)

wheirman:
> 1. Does `epoch_year = 1991.25` need to be changed to `2000.0`? 

https://cdsarc.u-strasbg.fr/ftp/cats/I/239/ReadMe mentions _Note on RAdeg, DEdeg: right ascension and declination are expressed in degrees for epoch J1991.25_ for `tyc_main.dat`, same as for `hip_main.dat` so it think `epoch_year = 1991.25` is correct?

> 2. Have you verified that each of the `na_values` is actually used in Tycho?

Here is `df.info` for the full `tyc_main.dat`. Removing some of the `na_values` introduced `object` Dtype for some but it looks fine with `na_values` as-is.

```
Index: 1058332 entries,    1    13 1 to 9537   380 1
Data columns (total 8 columns):
 #   Column            Non-Null Count    Dtype
---  ------            --------------    -----
 0   magnitude         1055115 non-null  float64
 1   ra_degrees        1058332 non-null  float64
 2   dec_degrees       1058332 non-null  float64
 3   parallax_mas      1035445 non-null  float64
 4   ra_mas_per_year   1035445 non-null  float64
 5   dec_mas_per_year  1035445 non-null  float64
 6   ra_hours          1058332 non-null  float64
 7   epoch_year        1058332 non-null  float64
dtypes: float64(8)
memory usage: 72.7+ MB
```

PR #973, opened 2024-06-10
introduce MDASegment for SPK type 1 and 21

### LINKED ISSUE #157: Unable to use .bsp files created by Horizons
I have been using the Horizons telnet interface to generate binary SPK files for small solar system bodies. When I try to load them in skyfield I get this error:

```
File "skyfield\jpllib.py", line 179, in __new__
    .format(spk_segment.data_type))

ValueError: SPK data type 1 not yet supported segment
```

which seems to be referring to the data types discussed here:
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/spk.html#Supported%20Data%20Types

Horizons is able to produce SPK files in a few different formats. The extended help for the file type selection prompt says this:

```
  [B]inary ... produce BINARY SPK file (.bsp). RECOMMENDED.
  [T]ext   ... produce ASCII transfer file (.xsp). DEPRECATED.
  1        ... produce legacy Type 1 binary format SPK file. DEPRECATED.
  -        ... move back to previous prompt
  ?        ... display a brief help file for this prompt
  ?!       ... display this extended help file
  X or Q   ... immediately disconnect from Horizons system

  * The initial default SPK format is "BINARY".  It is recommended this option
    be used.
```
...
```
  * The '1' response produces a Type 1 binary SPK file for use with older
    software linked against version of the SPICE Toolkit library prior to
    N0065. It is recommended users update to more recent SPICE reader libraries
    and not use this legacy option.

  * The alternate ASCII text form is deprecated and no longer necessary for
    users who link their software against N0052 and later version of the SPICE
    Toolkit reader library. However, if selected anyway, users must then
    convert the resulting ASCII format file to binary by running the SPICE
    utility program 'tobin', or 'spacit' on their local machines. Note that
    Horizons first produces the file in binary format then converts it to
    the ASCII transfer format if requested.
```
This seems to be implying that there are regular binary SPK files, and also legacy type 1 binary SPK files, but files created with either of these options all produce the same error I included above. I've tried this for a number of the largest small bodies (ceres, pallas, eris, etc.), and Horizons seems to only be capable of producing type 1 SPK files.

Based on a look through skyfield.jpllib.py it seems that skyfield currently supports data types 2 & 3. 

The problem could be that I'm using Horizons incorrectly, but if it's not are there any plans to support data type 1 SPK files? That would allow skyfield to be used for asteroids, dwarf planets other than pluto, and comets.

Thanks!

P.S. Another small issue is that after failing to load a type 1 SPK file, I tried to rename it and I got a dialog saying "The action can't be completed because the file is open in Python." Is the file not being properly closed if that exception is raised?

Tontyna:
Resolves #157 

Injecting the `MDASegment` into `skyfield.jpllib.SpiceKernel` is ... just a workaround. Might/should become a part of `jplephem.spk import SPK`

### DIFF
--- a/skyfield/jpl_mdasegment.py
+++ b/skyfield/jpl_mdasegment.py
@@ -1,0 +1,311 @@
+# coding: utf-8
+# jpl_mdasegment.py
+"""A supporting module for for skyfield to handle SPK data type 1 (Modified
+Difference Arrays) and type 21 (Extended Modified Difference Arrays)
+
+You can get SPK files for many solar system small bodies from HORIZONS
+system of NASA/JPL. See https://ssd.jpl.nasa.gov/horizons/
+
+At the point of June. 2024, HORIZONS system creates files of type 21 for
+binary SPK files by default.
+You can get type 1 binary SPK file for celestial small bodies through TELNET
+interface by answering back '1' for 'SPK file format'.
+
+Author: Tontyna
+
+This module has been developed based on Shushi Uetsuki's (whiskie14142) translation
+of the FORTRAN source of the SPICE Toolkit of NASA/JPL/NAIF.
+Its a extract and a combination of his Python modules spktype21 and spktype01.
+
+skyfield : https://pypi.org/project/skyfield/
+jplephem : https://pypi.org/project/jplephem/
+spktype21 : https://pypi.org/project/spktype21/
+spktype01 : https://pypi.org/project/spktype01/
+SPICE Toolkit for FORTRAN : http://naif.jpl.nasa.gov/naif/toolkit_FORTRAN.html
+SPK Required Reading : http://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/spk.html
+
+"""
+
+from numpy import array, zeros, reshape, searchsorted
+
+from jplephem.descriptorlib import reify
+from jplephem.spk import BaseSegment, T0, S_PER_DAY
+from jplephem.exceptions import OutOfRangeError
+
+class MDASegment(BaseSegment):
+    """Class for SPK kernel to handle data types 1 and 21 (Modified Difference Arrays)
+    """
+    def __init__(self, daf, source, descriptor):
+        super(MDASegment, self).__init__(daf, source, descriptor)
+
+        if self.data_type == 1:
+            self.MAXDIM = 15
+            # self.DLSIZE = 71
+        elif self.data_type == 21:
+            # type 21 has a difference line size (DLSIZE) indicating the length
+            # of the MDA records in the range of
+            #     [71 : (4*MAXTRM) + 1]
+
+            # Included from 'spk21.inc' on the FORTRAN source 'spke21.f'
+            MAXTRM = 25
+
+            # 'SPK Required Reading' indicates that the penultimate element of the segment
+            # is the difference line size (DLSIZE), but actual data contains there a MAXDIM.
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
I also haven't had time yet to read through this pull request, but I'll try to find time this month or next, so that I can respond before summer is fully over. Before a feature like this could become part of Skyfield, I'd need a sample file that it could be shown to work successfully on, and also documentation and tests. Also, could you describe the audience who will be using this, so we can think about where in the docs it would be described? Thanks!

Tontyna:
There is already a chapter about [Type 1 and Type 21 ephemeris formats](https://rhodesmill.org/skyfield/planets.html#type-1-and-type-21-ephemeris-formats) in the documentation. It recommends to install Shushi Uetsuki's [spktype01](https://pypi.org/project/spktype01/) resp, [spktype21](https://pypi.org/project/spktype21/) and implement a helper class. 
The `MDASegment` is based on those libraries -- I'd never managed to translate the FORTAN source of the SPICE Toolkit into Python :smirk:

For 'small bodies' like Chiron, Pholus and other not-so-popular asteroids there are no pre-built binary SPK files on the JPL- or NAIF-Server. 
A comfortable way to generate them is via the [HORIZONS Web Application](https://ssd.jpl.nasa.gov/horizons/app.html/)

Those *Small-Body SPK files* will most probably be

- of Type 21 
- heliocentric

Example code

```python
from skyfield.api import load

ts = load.timescale()
t = ts.utc(2024, 7, 12)

chiron_kernel = load('20002060.bsp')
planets_kernel = load('de440.bsp')

earth = planets_kernel['earth']
sun = planets_kernel['sun']

# add the sun to make it barycentric
chiron = sun + chiron_kernel[20002060]

ra, dec, distance = earth.at(t).observe(chiron).radec()
print(ra)
print(dec)
```
```
01h 23m 47.83s
+10deg 01' 20.9"
```

Tontyna:
Depending on the time span you selected the ephemeris file will be **multi-segmented**.
Unless you're only interested in the last one, you must explicitly pick the appropriate segment.

Instead of `chiron = sun + chiron_kernel[20002060]` do sth. like this:

```
def pick_segment(kernel, target, time):
    target = kernel.decode(target)
    tdb = time.tdb
    segments = kernel.segments
    # last segment should win
    segments = segments[::-1]
    for segment in segments:
        if (segment.target == target and
            segment.spk_segment.start_jd <= tdb and tdb < segment.spk_segment.end_jd):
            return segment

    format_datetime = '{0}-{1:02}-{2:02} {3:02}:{4:02}:{5:02.0f}'.format
    raise ValueError('kernel {0!r} has no segment for target "{1}" at TBD {2}'
                        .format(
                        kernel.filename,
                         _jpl_name(target) ,
                        format_datetime(*calendar_tuple(tdb))
                    ))

chiron = sun + pick_segment(chiron_kernel, 20002060, t)
```

Admittedly, without a 'MultiKernel' (cf.   #975) it's easier to use the **spktype21** library mentioned above. It picks the time-segment automatically.

Tontyna:
@brandon-rhodes As said before, I think the proper place for the `MDASegment` is  jplephem/spk.py. Shall I close this PR and move it tho jplephem instead?

Tontyna:
As for tests:
Requires at least one MDA-SPK file. Is there a recommended max. size for data files?

PR #975, opened 2024-06-16
Support SPICE ephemeris with multi-segments

### LINKED ISSUE #691: Support ephemerides (like DE441) that have 2 segments per planet
Hi,
sorry if I come back to the problem but I have a doubt. 
**de441.bsp: range date is wrong** https://github.com/skyfielders/python-skyfield/issues/682

If i use the files
- de441_part-1.bsp (1.6GB+)
- de441_part-2.bsp (1.6GB+)

downloaded from [https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/](https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/)
Skyfield has no problems.

If i use the file
- de441.bsp (3.2GB+)

downloaded from [ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/](ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/)
Skyfield returns this error:
```
Exception has occurred: EphemerisRangeError
ephemeris segment only covers dates 1969-07-29 23:59:18Z through 17191-03-14 23:58:51Z UT
```

But shouldn't it be the same?

Thanks a lot,
Blu

Tontyna:
resolves #691

`MultiKernel` can  load multiple `SpiceKernels` and combines their segments.
When a target's position is queried,  the appropriate `SPICESegments` are picked  from all segments available in all loaded SPK files.

Per definition that includes support of ephemerides that have more than 1 segment per planet.

### DIFF
--- a/skyfield/jpl_multikernel.py
+++ b/skyfield/jpl_multikernel.py
@@ -13,7 +13,7 @@ from numpy import array
 from jplephem.names import target_name_pairs
 
 from skyfield.api import load
-from skyfield.timelib import Time, calendar_tuple
+from skyfield.timelib import calendar_tuple
 from skyfield.vectorlib import (
     VectorSum, VectorFunction,
     _jpl_code_name_dict, _jpl_name
--- a/skyfield/jpl_multikernel.py
+++ b/skyfield/jpl_multikernel.py
@@ -1,0 +1,412 @@
+# coding: utf-8
+# jpl_multikernel.py
+"""
+A kernel capable of loading multiple SpiceKernels and combinig their segments.
+When a target's position is queried, MultiKerne/MultiSegment picks the appropriate
+SPICESegments from all segments available in all loaded SPK files.
+
+Per definition that includes support of ephemerides that have more than 1 segment per planet
+"""
+from collections import namedtuple, defaultdict
+from numpy import array
+
+from jplephem.names import target_name_pairs
+
+from skyfield.api import load
+from skyfield.timelib import Time, calendar_tuple
+from skyfield.vectorlib import (
+    VectorSum, VectorFunction,
+    _jpl_code_name_dict, _jpl_name
+)
+from skyfield.jpllib import (
+    _center,
+    _jpl_name_code_dict, _format_code_and_name
+)
+
+BaseChainTuple = namedtuple('BaseChainTuple', 'start end center chain')
+
+class MultiKernel(object):
+
+    def __init__(self):
+        self.kernels = []
+        self.kernel_segments = []
+        self.codes = set()
+        # collect attached MultiSegmenta for notifications
+        self._multisegments = []
+
+    def __repr__(self):
+        return '<{0}>'.format(type(self).__name__)
+
+    def __str__(self):
+        kernels = self.kernels
+        lines = ['SPICE multi segment kernel with {0} SPK kernel(s)'
+                 .format(len(kernels))]
+        for k in kernels:
+            lines.append(k.__repr__())
+        return '\n'.join(lines)
+
+    def __getitem__(self, target):
+        """Return a new MultiSegment object for computing the location of `target`."""
+        target = self.decode(target)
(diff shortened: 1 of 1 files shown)


Tontyna:
SPK files for celestial small bodies tend to be multi-segmented. But ... they probably consist of type 21 segments which aren't supported in Skyfield (yet).

Tontyna:
Test multi-segmented DE441
- segment 1:  JD -3100015.50 - JD 2440432.50  (-13200-05-06 through 1969-07-29)
- segment 2:   JD 2440432.50 - JD 8000016.50  (1969-07-29 through 17191-03-14)

```python
from skyfield.api import load
from skyfield.jpl_multikernel import MultiKernel

ts = load.timescale()
kernel = MultiKernel()
# it's possible to create ` MultiSegments`  in advance
earth = kernel['earth']
mars = kernel['mars']   # thanks to `MultiSegment._fallback_target`  this will work

kernel.load_kernel('de441.bsp')

# a time covered by first segment
t = ts.tdb_jd(1440432)
barycentric = earth.at(t)

# a time covered by second segment 
t = ts.tdb_jd(3440432)
barycentric = earth.at(t)

## critical tdb: obeserve() requires mars-vectors from both segments
t = ts.tdb_jd(2440432.501)
barycentric = earth.at(t)
astrometric = barycentric.observe(mars)
```

Tontyna:
The fact that `MultiSegment.center` is detected on the fly instead of being fixed right from the beginning has its drawbacks.

E.g. when a MultiSegment is involved in a `VectorSum` (by addition or subtraction)  Skyfield checks and combines the VectorFuntions' centers and targets.
No problem with the target, but the `VectorSum.center` determines the ICRF (sub)class used in subsequent calculations.

Without major changes to existing code, the only solution I can think of, is: 

1. assume that the center for the target will always be the same  in the time range covered by a given set of SPK files
2. call `MultiSegment.at()` to initialize its center

For example topocentric coordinates:

```python
from skyfield.api import load, N, W, wgs84
from skyfield.jpl_multikernel import MultiKernel

ts = load.timescale()
t = ts.now()

eph = MultiKernel()
eph.load_kernel('de441.bsp')

earth = eph['earth']
mars = eph['mars']

# !important: initialize earth's center before adding
earth.at(t)

boston = earth + wgs84.latlon(42.3583 * N, 71.0636 * W)
astrometric = boston.at(t).observe(mars)
alt, az, d = astrometric.apparent().altaz()

print(alt)
print(az)
```

brandon-rhodes:
@Tontyna — I'm not sure I'll have time this month to respond in detail, but I at least want to say, thanks for your interest in getting Skyfield set up to handle ephemerides that have two or more segments per target; it's definitely a use case I want to see added to Skyfield this year. I'm not sure I'll want to merge a solution that uses a `for` loop over time, though, as that's something I've been careful to avoid so far? The questions you raise about how the classes should be designed is a very important one, though, regardless of what kind of loop powers the computation under the hood; I plan to read over your solution carefully as I think through what direction the classes should take to add this ability to Skyfield.

Tontyna:
Of course, looping over time in a for-loop, isn't great. But what else is there to find the segment suitable for the requested time? Of course, a `pick-the-right-segment_for(t)` could be built into the SPICESegment, but somewhere somebody has to pick. And the neither the plain SPICESegment nor the single-file SpiceKernel know how to build the heliocentric chain when the required data  stem form multiple SPK files.

BTW.: My major intention for this MultiKernel wasn't the two-segmented DE441. My pet project is an astrological application, so I wanted a class that automatically builds the proper heliocentric VectorFunction chains for various asteroids (some of them  in multi-planet-multi-segmented SPK files, some of them of Typ 21). Covering a time span as big as possible.

Tontyna:
Anyway, there is no need to merge this single `jpl_multikernel.py` file into skyfield. If required, a plain local import will do.

PR #977, opened 2024-06-18
Resolve AttributeError in satellite_download.py 

jamescoledesign:
- It looks like an underscore is needed before the `exists()` method on the example for downloading element sets from [this webpage](https://rhodesmill.org/skyfield/earth-satellites.html). 
- Running the code from the example throws an error: `AttributeError: 'Loader' object has no attribute 'exists'. Did you mean: '_exists'?`
- This PR changes `exists()` to `_exists()` in satellite_download.py

### DIFF
--- a/examples/satellite_download.py
+++ b/examples/satellite_download.py
@@ -6,5 +6,5 @@ name = 'stations.csv'  # custom filename, not 'gp.php'
 base = 'https://celestrak.org/NORAD/elements/gp.php'
 url = base + '?GROUP=stations&FORMAT=csv'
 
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
@jamescoledesign — That example is, alas, only targeted at the latest version of Skyfield, in which that method was renamed to remove the underscore. Try upgrading to the latest version, and, I predict, the example will start working for you!

jamescoledesign:
Hi @brandon-rhodes I apologize for my delayed reply on this PR. I didn't realize I was using an older version—thanks for the clarification!

PR #979, opened 2024-06-25
Use arctan2 instead of arcsin when converting to spherical coordinates

wedesoft:
This pull request avoids using the _AVOID_DIVIDE_BY_ZERO term in two cases.

### DIFF
--- a/skyfield/functions.py
+++ b/skyfield/functions.py
@@ -2,7 +2,7 @@
 
 from numpy import (
     arcsin, arctan2, array, cos, einsum, finfo, float64,
-    full_like, load, nan, rollaxis, sin, sqrt, where,
+    full_like, hypot, load, nan, rollaxis, sin, sqrt, where,
 )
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Wow! This pull request is a surprise. I had always imagined `arctan2()`, with its multiple arguments, being much slower than `arcsin()`. And the call to `hypot()`, which I had thought might involve an intermediate Python call, in fact looks to be implemented in C. With the result that this patch seems to only incur about a 1% speed penalty for the routine, while eliminating the use of a somewhat awkward constant.

I'm inclined, therefore, to merge. I'll wait till tomorrow in case any edge cases occur to me, but I think this is likely to land. Thanks!

wedesoft:
Ok great. I think atan2 is numerically more stable than arcsin especially for angles close to +90 or -90 degrees where sinus is near 1.0 or -1.0.

PR #984, opened 2024-07-20
Remove unused modules

AlexKurek:


brandon-rhodes:
@AlexKurek — Could you provide a bit of background on your goals in opening this pull request, and the similar requests in other repositories?

image

AlexKurek:
@brandon-rhodes I was just making progress in Python by looking at other codes and linters like Ruff or Pylance. The idea was to do it in a way that would hopefully be useful for astrophysics. I find more issues, but these are relatively safe and easy to offer a fix for.

brandon-rhodes:
Understood — so these are not changes that you yourself need to complete your work, but were suggested to you by an automated tool.

In this case I won't be applying the changes, as Python files in `design/` are more like notepads where I try out ideas, and I usually don't prune their `import` statements because a function that I've commented out or aren't using for the moment is likely to come back as I do more editing later. Skyfield already has CI tests to make sure imports are kept minimal where it matters, in Skyfield's own code.

Issue #959, opened 2024-04-28
Comets and Minor Planets throw ValueError with a narrow date in find_risings()


Bernmeister:
Using a date range of 24 hours to find risings (presumably too find settings ) for comets and minor planets throws a `ValueError`.  See the example below.  To test, comment/uncomment the `t0` / `t1` pairs.

The "fix" was simple: increase the date range to 48 hours.  

```
#!/usr/bin/env python3


from skyfield import almanac
from skyfield.api import N, S, E, W, load, wgs84
from skyfield.constants import GM_SUN_Pitjeva_2005_km3_s2 as GM_SUN
from skyfield.data import mpc

ts = load.timescale()

eph = load('de421.bsp')
sun = eph['Sun']
earth = eph['earth']

bluffton = wgs84.latlon(40.8939 * N, 83.8917 * W)
observer = eph['Earth'] + bluffton

# FAILS
t0 = ts.utc(2024, 4, 26, 4)
t1 = ts.utc(2024, 4, 27, 4)

# PASSES
# t0 = ts.utc(2024, 4, 26, 4)
# t1 = ts.utc(2024, 4, 28, 4)


with load.open(mpc.COMET_URL) as f:
    comets = mpc.load_comets_dataframe(f)

comets = (comets.sort_values('reference')
          .groupby('designation', as_index=False).last()
          .set_index('designation', drop=False))

halley = sun + mpc.comet_orbit(comets.loc['1P/Halley'], ts, GM_SUN)

# ValueError: non-broadcastable output operand with shape (3,1) doesn't match the broadcast shape (3,3)
t, y = almanac.find_risings(observer, halley, t0, t1)
print(t.utc_iso(' '))
print(y)


with load.open('MPCORB.excerpt.DAT') as f:
    minor_planets = mpc.load_mpcorb_dataframe(f)

minor_planets = minor_planets.set_index('designation', drop=False)

ceres = sun + mpc.mpcorb_orbit(minor_planets.loc['(1) Ceres'], ts, GM_SUN)

# ValueError: non-broadcastable output operand with shape (3,1) doesn't match the broadcast shape (3,3)
t, y = almanac.find_risings(observer, ceres, t0, t1)
print(t.utc_iso(' '))
print(y)
```

Tontyna:
Reason that the narrow dates fail is that there is only one rising inbetween.

[...]

PR #986, opened 2024-07-23
fix shape returned from keplerlib.propagate

Tontyna:
fiixes #959

### DIFF
--- a/skyfield/keplerlib.py
+++ b/skyfield/keplerlib.py
@@ -5,7 +5,7 @@ import math
 from numpy import (
     abs, amax, amin, arange, arccos, arctan, array, atleast_1d,
     clip, copy, copyto, cos, cosh, exp, full_like, log, ndarray, newaxis,
-    pi, power, repeat, sign, sin, sinh, squeeze, sqrt, sum,
+    pi, power, repeat, reshape, sign, sin, sinh, squeeze, sqrt, sum,
     tan, tanh, zeros_like,
 )
 from skyfield.constants import AU_KM, DAY_S, DEG2RAD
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
@JoshPaterson — I'm inclined to accept this pull request, maybe after requesting that a test be written so that a future well-meaning coder doesn't break this edge case again. But I thought I should loop you in, since I think this is your code? Simply comment here over the next few days if you'd like to weigh in. Otherwise, if I don't hear from you by maybe the middle of next week, I'll assume you're busy these days. I hope you're doing well!

JoshPaterson:
Hi @brandon-rhodes, I'll take a look at this. I'm wondering if instead of checking if squeezing will make the wrong shape and then using a conditional to fix it, can the squeezing be made to work correctly the first time? If either you or @Tontyna have ideas about that let me know!

JoshPaterson:
Perhaps this fix can be accepted to get things working and then if it can be improved that can be a separate PR.

Tontyna:
I tried to avoid reshape, but squeeze can't do nothing about it. 
The whole calculation -- `atleast_ld` ! -- doesn't distinguish between a given scalar time and a [time], so we must remember (and reshape).

Tontyna:
One could do without the `expected_shape` variable by not modifying the `t1` parameter within the function and doing the shape-check / reshape only at the end of the function.

brandon-rhodes:
I came up with a slightly simpler solution:

fb120742b5409aa90e3d723b375e3ccd1fd8fe26

So I'm going to close this PR in favor of my little fix. But thanks for getting a solution up online that folks could apply in the meantime!

PR #1014, opened 2024-10-29
Fix #1012 DeprecationWarning: datetime.datetime.utcnow() is deprecated

### LINKED ISSUE #1012: DeprecationWarning: datetime.datetime.utcnow() is deprecated
Skyfield is using `utcnow()` which is being deprecated:
https://github.com/skyfielders/python-skyfield/blob/2d104a43b47ad37e783038ed96fb8048132fb778/skyfield/timelib.py#L134

This results in warnings being thrown:

```
.venv/lib/python3.12/site-packages/skyfield/timelib.py:134: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    return self.from_datetime(self._utcnow().replace(tzinfo=utc))
```

ruitcatarino:
Fixes #1012 
Refactored the code to use `datetime.now(timezone.utc)` instead of the deprecated version `datetime.utcnow`. 

https://docs.python.org/3/whatsnew/3.12.html
> datetime: datetime.datetime’s utcnow() and utcfromtimestamp() are deprecated and will be removed in a future version. Instead, use timezone-aware objects to represent datetimes in UTC: respectively, call now() and fromtimestamp() with the tz parameter set to datetime.UTC. (Contributed by Paul Ganssle in gh-103857.)


brandon-rhodes:
@ruitcatarino — Happy New Year! I wanted to check back in and ask about the test failures shown above. Are you likely to have time to address them and get the build green for this PR? Or are there more moving parts here than you have time to address?

ruitcatarino:
Hi @brandon-rhodes,

Thank you for taking the time to review this PR.

Upon reviewing the pipeline [tests for Python 2](https://github.com/skyfielders/python-skyfield/actions/runs/11572106248/job/32594609999?pr=1014), I noticed that the library still supports Python 2. Since Python 2 does not include the timezone class (which is available only from Python 3.2 onwards), we need to address this for compatibility. (Reference: [Python 3 datetime module - timezone objects](https://docs.python.org/3/library/datetime.html#timezone-objects))

To maintain backward compatibility with Python 2, we could conditionally use the timezone class for Python versions 3.2 and above while retaining the previous behavior for earlier versions. This approach will ensure that the library continues to function correctly across all supported Python versions.

Let me know your thoughts on this approach or if you have any suggestions for handling this differently.

brandon-rhodes:
Let's take a look. Hmm. Ah, check this out! The `timelib.py` module already has a symbol `utc` that looks similar to what you are talking about:

https://github.com/skyfielders/python-skyfield/blob/master/skyfield/timelib.py#L50

Would it be possible for you to use that symbol, or does it not quite fit what you need?

ruitcatarino:
Hi @brandon-rhodes,

Wow, I can't believe I missed that existing implementation of `utc`. Thanks for pointing it out!
Given this, I have updated the code to replace the `_utcnow` with a direct call to `datetime.now(utc)`. This simplifies the code by removing the unnecessary abstraction.

However, I noticed that the doctest is failing, which suggests that the implementation might not be totally equal. I’d appreciate your thoughts on the changes I’ve made and any guidance on how to resolve the doctest issue.

Looking forward to your feedback!

brandon-rhodes:
> Wow, I can't believe I missed that existing implementation of `utc`.

It's kind of awkwardly positioned in the file, so it probably didn't stick out at first reading!

> However, I noticed that the doctest is failing, which suggests that the implementation might not be totally equal. I’d appreciate your thoughts on the changes I’ve made and any guidance on how to resolve the doctest issue.

Because doctests need to be stable, and always return the same output, they can't really call the real `ts.now()` routine, because it gives a different date and time on every call. Instead, each doctest that uses `.now()` calls `skyfield.tests.fixes.setup()` at the top, like this:

https://github.com/skyfielders/python-skyfield/blob/master/documentation/examples.rst?plain=1#L23

The argument specifies what `.now()` should return. Or, there's a default if the test suite doesn't want to specify one:

https://github.com/skyfielders/python-skyfield/blob/master/skyfield/tests/fixes.py#L19

It looks like your pull request currently disables that mechanism, but without putting another one in its place? Let me know if you think I should add a comment or two in `timelib.py` explaining what's going on; my editor makes it easy for me to ask "What's this `._utcnow()` nonsense? Who uses that?" and jump immediately to all the other uses of that symbol, to remind myself that it's used in testing. But maybe I shouldn't assume that other folks can quickly gain context like that, and just put in a `# for testing; see tests/fixes.py` next to it?

brandon-rhodes:
Since I didn't hear anything further, I went ahead and wrote up a slightly different approach, that preserves the behavior of the doctests:

c2e5c4ab8052f4a76cbb8de11143f9f3f91eea5c

Thanks for the help, though; I know it's tough to get back to a PR when it turns out more work is involved.

PR #1076, opened 2025-04-20
Add routine and test for the solar eclipse calculation.

mindaugl:
Adding routine and test for finding solar eclipses.
It uses similar method as lunar eclipse calculation, but instead of looking for maximum of angle between earth-sun and earth-moon, it looks for minimum.
Solar eclipse routine also:
- Identifies solar eclipse type (Total and Hybrid are grouped together, because it would require to look at multiple time instances around eclipse time to differentiate the two).
- Calculates magnitude and obscurity for each eclipse.

The agreement of routine results with the official data for the last ~400 years is rather good.
E.g., the few misidentifications of eclipse type are for the eclipses that last 0 seconds.
The few eclipses that are not found or found extra, are for low magnitude partial eclipses, where, e.g., using polar earth radius versus equatorial, would make the difference for the eclipse identification.

### DIFF
--- a/skyfield/eclipselib.py
+++ b/skyfield/eclipselib.py
@@ -3,10 +3,10 @@
 
 from __future__ import division
 
-from numpy import arcsin, byte
+from numpy import arcsin, byte, arctan2, sin, pi, arccos, sqrt, dot, array, empty
 from .constants import AU_KM, C_AUDAY, ERAD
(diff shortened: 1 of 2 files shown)


mindaugl:
More details of the change and comparison of results to published eclipse can be found in the Issue #1078.

